PR #2, opened 2022-12-08
Enable deflate compression for fsdata.c

deeebug:
Hi,

I've compiled new binaries for `makefsdata` with `miniz` enabled. This should significantly reduce the size of `fsdata.c` because this build of `makefsdata` uses deflate.

This reduced the size of `fsdata.c` from 4MB to 1MB. And that has allowed the reduction of the size of UF2s significantly!

Please verify and accept the merge if alright.

Thanks!


arntsonl (approved):
Looks good, we should revisit how makefsdata and makefsdata.exe are built in the future from the lwip project.

PR #3, opened 2022-12-09
Add list of PS4 fighting games to FAQ

StdVectorBool:
Pulled from various Reddit threads.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -10,6 +10,24 @@ Generally speaking, XInput will be the mode of choice for everything except Nint
 
 GP2040 will work on PS4 games that implement support for legacy PS3 controllers. Many of the popular PS4 fighting games have this support.
 
+Supported:
+* BlazBlue Central Fiction
+* Guilty Gear Xrd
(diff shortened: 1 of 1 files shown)


henrebotha:
I think it would be better not to list games that are _not_ compatible, because that gets us into "maintain a database of all PS4 fighting games" territory. We could restrict it only to a list of games we've checked, and add a very clear caveat that we could be wrong, it might have changed since we checked, etc.

FeralAI:
I believe there is value in having both confirmed lists (working/non-working) without users needing to search the internet for various threads to confirm. The games we're referring to here are a generation out and there won't be many new entries to capture once a stable list is built. It does not have to be a live document and can receieve sporadic updates and PRs.

I've recently discovered several of the M2 ShotTriggers shmup releases on PS4 support legacy input and would be great to capture.

If it's agreed to be out of scope for this repo, I propose considering a stanalone repo with just the compatibility lists.

henrebotha:
I agree with sticking this in a standalone project.

PR #5, opened 2022-12-12
Fix PS button for PS3

deeebug:
Hi!

This is a companion fix to be applied with https://github.com/OpenStickCommunity/MPG/pull/1

This fixes the call made in response to supply the magic bytes and also references latest MPG which is to be merged with aforementioned fix.


arntsonl (approved):
This is perfect, moves our MPG to our community version and fixes the tud_hid code for PS3 legacy.

PR #8, opened 2022-12-19
Add A1-PS map info to board configs comments/docs

InfraredAces:


troytroy22:
Also remove button 14 from list in ps3 mode.

arntsonl (approved):
Nice little update

PR #20, opened 2022-12-21
Changes to support the CMake build system

deeebug:
This PR migrates the build system to CMake and is a heavy change on how the sources are built and how deployments will be handled.

All heavy lifting done by @Fortinbra!
It's a massive update that the project was in need of.
I added some entries to get the CMake build working and edited a file to register the PS button callback. 
Edit: Also, added the web-config fix for macOS.

Please verify and merge this.

Thanks!
Again, massive thanks to @Fortinbra for this great move!


deeebug:
Alright, also added the web-config fix for macOS which this update enables.

SuperPrower:
Hi, is this ready to be used for builds, for testing purposes? How one would go about specifying which board preset to use? Can't seem to find anything about that in CMake files, except just copying include flags from the old pio config files

Issue #1, opened 2022-12-06
Migrate to CMake build system, instead of PlatformIO


Fortinbra:
Migrate away from PlatformIO and the Wizio-Pico dependencies, and use the Pico SDK directly via CMake.

Will need to adjust how we approach builds for community boards.

PR #21, opened 2022-12-21
Add passive buzzer speaker support

rafaelcruzpb:
Add support for the passive buzzer speaker, with the possibility of playing various tones and even creating music (programmatically for now)


arntsonl:
That is a cool idea!! Are you in our discord channel? https://discord.gg/k2pxhke7q8   I'm lucipher, one of the mods on there.

I have some suggested changes and some code you can use to make this functional (On Device Connected) before we merge this into Main.

arntsonl (changes requested):
#1 your web-config changes should be in the AddOns page rather than an entirely new page. I'm going to be adding an Enabled [x] flag to that page soonish, so all of our add-ons should be there

#2 your enum Tone can live in BuzzerSpeaker.cpp since none of the other files will use it. that helps encapsulate your defines inside of that file rather than having it defined everywhere

#3 processBuzzer, play, and stop can be private functions of BuzzerSpeakerAddon unless those functions need to be exposed as part of their required usage. it looks to me like process() can handle events so for now we can keep those hidden


and then for actual functional code for your buzzer: check out the code in here https://github.com/OpenStickCommunity/GP2040-CE/blob/main/src/addons/board_led.cpp  from @debug, there's a USB mounted event you can poll from to play a song. that would be a good functional usage of this add-on

rafaelcruzpb:
@arntsonl These changes have been made.

About point #2 , Tone struct has dependency on .h file itself, and on songs.h file as well

arntsonl (approved):

PR #37, opened 2023-01-04
PicoAnn device configuration

TheTrainGoes:
This adds the PicoAnn original pinout for GP2040-CE which is used on the Pico Advanced Breakout Board v3.0


arntsonl (approved):
Looks good, this will give us a link for v3 users which is nice.

PR #40, opened 2023-01-04
Fix board LED to account for all inputs with raw input state.

deeebug:
This makes the LED react to unprocessed input state.
Please verify and merge.

Thanks!


arntsonl (approved):
Seems like a simple fix!

PR #44, opened 2023-01-05
Add OLED display saver feature

rafaelcruzpb:
Adds a configuration option that allows you to define a time in minutes referring to the control inactivity timeout, and then the oled display will be turned off.

This is useful for those who always keep the controller connected to the computer, and some computers keep the USB powered when they are turned off.


bjszd:
Why is fsdata.c changed in this pr?

rafaelcruzpb:
@bjszd Because there were changes in the "src/Pages/DisplayConfig.js" file of the web-config submodule. These changes are in this PR: https://github.com/OpenStickCommunity/WebConfigurator/pull/6

arntsonl (changes requested):
The top of headers/addons/i2cdisplay.h needs a define for DISPLAY_SAVER_TIMEOUT when it is not defined in boardconfig.h

Right now it will throw an error if you try to compile boards other than Pico, so you'll have to make sure you do something like

#ifndef DISPLAY_SAVER_TIMEOUT
#define DISPLAY_SAVER_TIMEOUT -1
#endif

otherwise storagemanager.cpp will not know what boardOptions.displaySaverTimeout is

rafaelcruzpb:
@arntsonl Oh right, I fixed that. Thanks

arntsonl (approved):
Looking good, going to run the build action and confirm it before I pull it in.

Issue #18, opened 2022-12-21
0.5.2 stick animation in DP is OK, if use LS the OLED not have any move on joystick animation


TheTrainGoes:
"0.5.2 version, I found OLED stick animation if I use DP is OK,
but if use LS model the OLED not have any move on joystick animation, just joystick ball on center not move"

I have tested this and it is due to LS and RS not being defined in the gamepad. Not sure if this can be fixed but we will look into it further

PR #45, opened 2023-01-08
Fixed bug #18

mthiesen:
This fixes the stick animation bug in the `DPAD_MODE_LEFT_ANALOG` and `DPAD_MODE_RIGHT_ANALOG` modes. The code now takes the current D-Pad mode into account when displaying a direction.

One could make the point that the unprocessed input should be displayed instead, but doing this would also disregard SOCD cleaning on the display. This leads to the more fundamental question of whether the display should show the actual input or what is send to the PC / console. I favoured the latter interpretation because I think it is more useful for the user.

### DIFF
--- a/headers/addons/i2cdisplay.h
+++ b/headers/addons/i2cdisplay.h
@@ -64,6 +64,7 @@ public:
 	virtual void setup();
 	virtual void process();
 	virtual std::string name() { return I2CDisplayName; }
+private:
 	void clearScreen(int render); // DisplayModule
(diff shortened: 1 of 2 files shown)


TheTrainGoes (approved):
Hi @mthiesen ,

Thank you for the PR to fix this.  

I have tested this and there are no issues that I can find on the base configuration.  

As a note, it appears to not work when the `reverse input` add-on is enabled and used.  I will make a separate ticket for this.

PR #46, opened 2023-01-10
Configs for normal and mirrored Fightboard v3

thnikk:
This adds configuration files for the two versions of the Fightboard v3.


TheTrainGoes (commented):
Hi Thnikk,

Changes all look good.  

Are you on Discord?  Wanted to ask you about a few things on the OLED layout.

TheTrainGoes (approved):
Conditional approval.  I don't have an MX to test this on.  

There are a few things that we should do to make this match the pictured MX:
- Match the aux buttons that are shown on the display (we can do a custom display layout for this)
- Look into the original display layout that uses angled WASD keys
- The flipped version has the wrong display layout (it as well will need custom layouts put together)

TheTrainGoes (approved):

PR #47, opened 2023-01-10
Switch int values to proper absolute_time_t type

deeebug:
Allows Debug build types to be built.

Please verify and merge.
Thank you.


arntsonl (approved):
Makes sense, getting rid of errors

PR #50, opened 2023-01-14
Fixed multiple debug build issues in FlashPROM

mthiesen:
Currently debug builds trigger multiple assertions in the Pi Pico SDK when `FlashPROM::commit()` is called.

This PR fixes thse issues:

`cancel_alarm(0)` will assert in `alarm_pool_cancel_alarm()` in `time.c`. I changed it so `cancel_alarm()` is only called when there actually is a pending alarm (`flashWriteAlarm != 0`).
 
Both `multicore_lockout_start_blocking()` and `multicore_lockout_end_blocking()` assert. Behind the scenes both functions actually use a timeout with a very large time constant (`0x7fffffffffffffff`). Unfortunately the time constant is so large that the subsequent calculations result in a `absolute_time_t` value outside of the allowed range, triggering the assert in `update_us_since_boot()`. I worked around this issue by using a smaller but still very long timeout (> 30,000 years).

### DIFF
--- a/lib/FlashPROM/src/FlashPROM.cpp
+++ b/lib/FlashPROM/src/FlashPROM.cpp
@@ -13,14 +13,19 @@ int64_t writeToFlash(alarm_id_t id, void *flashCache)
 {
 	while (is_spin_locked(flashLock));
 
-	multicore_lockout_start_blocking();
+	// We use a very long timeout (> 30,000 years) instead of multicore_lockout_start_blocking() because the latter
+	// asserts in debug builds.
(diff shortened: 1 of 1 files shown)


mthiesen:
A fix for the `multicore_lockout_xxx` issue is actually coming in Pico SDK 1.4.1 see: https://github.com/raspberrypi/pico-sdk/issues/1059.

arntsonl (approved):
Good catches on these! We can circle back with the fix from 1.4.1 but that's a good alternative. I want to pull out the PROM code so I might have you take a look at my new LittleFS chagnes.

PR #51, opened 2023-01-14
Fixed multiple compiler warnings

mthiesen:
I fixed compiler warnings that were at least triggered with my build setup (Arm GNU Toolchain 12.2.Rel1 (Build arm-12.24)). In more detail, this is what I did:

- Removed `typedef`s from `addonmanager.h`. These triggered the `'typedef' was ignored in this declaration` warning. `typedef`s are pretty much useless in C++ (as opposed to C) anyway.
- I fixed multiple instances of `warning: control reaches end of non-void function`. I added `assert(false)` where the code path should be unreachable. But at least in the case of `AddonManager::GetAddon()` this was an actual issue. Calling this function with an unknown addon name would cause undefined behaviour.

There are still some compiler warnings left in `httpd.c`.  I did not touch those as this is external code.

Also there are some linker warnings, but I think those are to be expected because the libc implementation of gcc-arm-none-eabi is incomplete.

### DIFF
--- a/headers/addonmanager.h
+++ b/headers/addonmanager.h
@@ -6,12 +6,12 @@
 #include <vector>
 #include <pico/mutex.h>
 
-typedef enum ADDON_PROCESS {
+enum ADDON_PROCESS {
     CORE0_INPUT,
     CORE1_LOOP
-} _ADDON_PROCESS;
+};
 
-typedef struct AddonBlock {
+struct AddonBlock {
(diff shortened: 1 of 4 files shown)


arntsonl (approved):
These look good, I'm not sure if we have an alternative to assert(false) for handling errors gracefully. I might circle back on logging errors but assert is fine for now.

Issue #17, opened 2022-12-21
Proof of concept input histroy.


TheTrainGoes:
There was a PR on the original GP2040 that had the foundation for input history from Sleep:

https://github.com/FeralAI/GP2040/pull/48

I compiled and tested this.

It currently is a running line that will max out on screen length and will only resister one input of each input type in a row. As an example, if I were to press left and then left again, it would only register one instance of left.

For this to be more useful we would want to limit the amount of space it takes up and also have it include additional instances of input types that are pressed in a row.

TheTrainGoes:
Update on this - As soon as #53 gets updated with WWW we can merge and close this.  Tested and working!  Cool little addon

TheTrainGoes:
This is finished and now in main

PR #53, opened 2023-01-15
Input history

thnikk:
Implements #17. This uses some of the code from the original [GP2040 PR](https://github.com/FeralAI/GP2040/pull/48), but correctly shows when the same button is pressed multiple times. I'm opening this as a draft because I think there are some things that need to be worked out first:

History picture


TheTrainGoes:
Great work @thnikk !

I checked this out and it is a step forward from the first proof of concept that we had on this.  

The two big things that stood out to me are that combo buttons are not represented and it does cut off the bottom line of most layouts.

TheTrainGoes:
Hi @thnikk - some changes have been made to www and will need to be updated for this to compile now.

thnikk:
This is almost ready but I've only tested it with the Fightboard layout so the others still need to be adjusted. This should be pretty straightforward so I'll try to get it done ASAP.

thnikk:
I've moved the layouts that fit but the dancepad and stickless layouts are both too big. The dancepad layout could probably be scaled down slightly without being too much smaller, but the stickless layout either needs the up button to be moved or the whole thing needs to be scaled down.

TheTrainGoes:
@thnikk happy to check this out tomorrow.  

I would prefer slight moves in position over resizing as most of the lays do not resize super well.  

If this is going to be added to the add-on section you could possibly add a second variable that says how long of a string to display for working around the stickless layout.

thnikk:
I've adjusted the last two layouts to make them fit. The dancepad layout wasn't much of a problem, I just lowered the button size a bit and moved it up. The stickless layout needed the up button to move up, which barely fits with no gap and changes the look of the layout a bit. 




I tried limiting string length for the original stickless layout, but I think it reduces the utility a lot with so little room on the left side. At that point, I think it would make more sense to only show the last key that was pressed, but even that would limit some 3 key combos from being displayed.

TheTrainGoes:
Hey @thnikk - compiled this locally and tested out.

Great progress on the actual input history part.  That is working very well and has addressed all of the original concerns.  

I agree that reducing the visible portion of input history will reduce its overall use and appeal so I am happy to see it running across the entire bottom of the screen.  

I would like to see this added as an add-on that has the ability to be toggled on and off.

Assuming that this is added as an add-on I would like to ensure we can keep original layouts and possibly just have modified layouts where needed just for the input history.  

Stickless is a prime example.  The current build does not look quite right and rather cramped.  Given that people don't spend all of their time looking at the screen I would feel that the reduced size buttons (while octagonal) are a better overall option when used with this add-on than to have a major layout change.  

The easy solution to this will just be to have an additional set of layouts that can be enabled if you want to use the add-on.  The harder solution would either be to have that automatically done based on the add-on being turned on or have an overriding set of options that are available only when you turn the add-on on.  

Happy to chat more about this on Discord!

arntsonl:
This seems like a very neat feature for the OLED, but we want to make sure we don't have any layout breaking/changes that require a lot of updates.

I'm going to close this PR until we have the ability to turn input history on and off through the web config. Maybe we could also have a hotkey or something similar to do this.

Once you have the ability to toggle this mode, create a new PR and we can get it merged in.

PR #54, opened 2023-01-15
Add button and splash display preview in config

deeebug:
This enables B1 and B2 button press to show either button layout or splash preview respectively.
Note that the screen only updates when save is clicked.


arntsonl (approved):
Display preview B1 > Button works great! 

B2 > Splash works once you click Enabled :) That's an interesting use case though, very helpful!

PR #55, opened 2023-01-15
Add feature to have customized button layout from web-config

deeebug:
This update builds upon the display preview feature.
From the web-config, for preset button layouts parameters can be changed and also be previewed instantaneously.

This is how the added section looks like:
image


arntsonl (changes requested):
Let's avoid changing setBoardOptions() to have a pseudo-change without commits.

A better solution would be BoardOptions previewOptions; with a deep copy of our current boardoptions. This prevents our Flash and RAM from going out of sync, which would be confusing and difficult to follow.

arntsonl (approved):
That looks much better! We should clean up the i2cdisplay at some point here but that takes care of not setting board options with temporary flags.

chiyu521:
C++ sorry, unimplemented: non-trivial designated initializers not supported

PR #56, opened 2023-01-15
Implement addon BOOTSEL button

deeebug:
This addon allows BOOTSEL button to be used as an extra input which should be good for diagnostics and quick checks of boards.
OLED may become unresponsive while this addon is enabled (i.e., if any button is set).

image


arntsonl (changes requested):
It looks like the OLED stopping is one of the issues.

The other issue is the Pico becomes non-responsive if you save and use the RUN pin on board. Hopefully fixing the OLED issue fixes the reboot issue.

arntsonl (approved):

Issue #15, opened 2022-12-21
Force SH1106/"FORCE" mode in the display settings


TheTrainGoes:
Add an option to force the SSH1106 / LCD options in the display settings. This will solve the edge case with certain SH1106 displays not showing our OLED code.

@thesuccubus current main fork supports this but should be changed:
https://github.com/thesuccubus/GP2040-CE

mthiesen:
It might actually be possible to auto-detect the display type. [FlashFloppy](https://github.com/keirf/flashfloppy/) does this. Look at the [oled_probe_model](https://github.com/keirf/flashfloppy/blob/feb0b883da32552035ce5292539d76162329c682/src/display/lcd_stm32f105.c#L1002) function.

thesuccubus:
SSH1106 screens sometimes misidentify themselves and cause issues as they are actually 132 px high and try to mask themselves as an SSD display, in my fix I have to force it to recognize it. But having the default iAddr set to -1 (auto recognition) might catch more display types and solve other issues.

mthiesen:
Sure, it is just that currently it seems like there is not even an attempt made to auto-identify the display. This is why I think it might be worth giving the FlashFloppy approach a try. In the end there should be a way to set the display type explicitly of course, to catch cases where this fails.

thesuccubus:
The library already being used already tries auto identification when the default iAddr is set to -1, it does already work if you set this variable manually. In one of my test forks I do use that. So just one change on the iAddr should work

mthiesen:
This does not seem to work, at least not with my SH1106, I still get the same horizontal offset in the display.

Looking at the code of `odbI2CInit()` one can see that passing `-1` as `iAddr` triggers the I2C address detection code in line 462-468. But beyond that `iAddr` is not even used during initialization, in particular not to detect a display type.

Irrespective of the value of `iAddr` the code does make an attempt to differentiate between SH1106, SH1107 and SSD1306 (lines 482-496). I have just stepped through this code with the picoprobe debugger and verified that it does not detect my display correctly. In fact it does not take any of the `if` branches.

The only thing that works for me right now passing `OLED_132x64` as `iType` to `obdI2CInit()`.

FlashFloppy seems to exploit the fact, that SH1106 does allow readback from the internal display memory over I2C, while SSD1306 does not. So they write some random bytes to the display and attempt to read them back, if it works the display is detected as an SH1106, otherwise it is assumed to be an SSD1306. I think this should work here as well.

[...]

PR #57, opened 2023-01-15
Auto-detect SH1106 OLED displays

mthiesen:
This is my attempt at solving #15.

I use the same general strategy as the [FlashFloppy](https://github.com/keirf/flashfloppy/) project for differentiating between SSD1306 and SH1106 displays. SH1106 allows read operations over I2C whereas SSD1306 does not. I do some read-modify-write operations and if I can read back data I previously wrote, I re-initialize the display as SH1106, otherwise I leave it as is.

I tested this with both types of displays and for me this approach works perfectly. Testing with a wider variety of OLEDs would be better, of course. I only own two makes of OLEDs.

I can see the merit of a user-configurable options as well, but I think this is a better default. A manual override option can be added to the web config at any time.

### DIFF
--- a/headers/addons/i2cdisplay.h
+++ b/headers/addons/i2cdisplay.h
@@ -65,6 +65,8 @@ public:
 	virtual void process();
 	virtual std::string name() { return I2CDisplayName; }
 private:
+	int initDisplay(int typeOverride);
+	bool isSH1106(int detectedDisplay);
 	void clearScreen(int render); // DisplayModule
 	void drawStickless(int startX, int startY, int buttonRadius, int buttonPadding);
 	void drawWasdBox(int startX, int startY, int buttonRadius, int buttonPadding);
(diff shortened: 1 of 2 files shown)


arntsonl:
Wow this is a much better solution. I'll have @TheTrainGoes and myself test on the OLEDs we have, but this is way better.

troytroy22:
I can confim this is working and fixed my 1.3 oled that had off center problem without issues.

TheTrainGoes:
Good to know this is working on one of the known issue displays.  Thanks @troytroy22 

I’ll test on my displays when I have a chance and if all is well with them I can merge.

TheTrainGoes:
@mthiesen had a chance to test this one - apologies on the wait.  

Working on both of my screens without issue so I am happy to get this merged.

TheTrainGoes (approved):
Changes look good.  Was tested on two of my displays (SSD1309 and SSH1106).  @troytroy22 was able to test on one of the SH1106 displays that had the wrong sizing on it and that worked as well.  

Approving and merging

PR #62, opened 2023-01-20
Fightboard LED theme

thnikk:
This adds the Fightboard LED theme and sets it as the default theme for Fightboards.


arntsonl (approved):
Neat little theme!

PR #63, opened 2023-01-20
Fightboard display layouts

thnikk:
This adds display layouts for both the regular and mirrored Fightboards.


TheTrainGoes (dismissed):
Looks good - We may change the way these are handled in web-config in the future

arntsonl:
@thnikk we're having a problem merging fsdata.c.

Since this is built by our web configurator, could you revert changes to /lib/httpd/fsdata.c and commit? Then we'll squash & merge this and build w/ your webconfigurator changes.

thnikk:
I'm not sure if my upstream merge was necessary but that should do it.

PR #64, opened 2023-01-21
Added some missing #include guards

mthiesen:
Just a small fix, I noticed that two headers were missing #include guards.


TheTrainGoes (approved):
Looks good - tested and working.  Thank you!

Issue #60, opened 2023-01-19
Can we add Memory Usage to Web-Config?


arntsonl:
I honestly don't know if we can or how we'd do it, but that would be neat.

It might not be possible outside of running a debugger on the chip, but the FlashPROM issue made me think about it.

If there's no way, we can close this issue.

arntsonl:
Looks like we can!! wow thanks @mthiesen

PR #65, opened 2023-01-22
Implemented memory report endpoint

mthiesen:
This is my attempt at implementing #60.

I implemented various memory-diagnostic functions in a new `System` namespace:

- `System::getTotalFlash()` - Returns the size of on-board flash memory
- `System::getUsedFlash()` - Returns the amount of on-board flash memory used by the firmware. This only includes flash memory used by the firmware, not the memory used by `FlashPROM`.
- `System::getStaticAllocs()` - Returns the amount of memory used for static allocations (e.g. static variables, static arrays, the stack for both cores, and code placed into memory for speed purposes)
- `System::getTotalHeap()` - Returns the total amount of memory used for dynamic allocations
- `System::getUsedHeap()` - Returns the amount of dynamically allocated memory

I also implemented an API endpoint for the Web Config Server. Navigate to [http://192.168.7.1/api/getMemoryReport](http://192.168.7.1/api/getMemoryReport) while in Web Config Mode to see the raw JSON data.

Of couse this will have to be properly integrated into the WebConfig project after this PR is merged. I am no webdev, so maybe someone else can do this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ src/gamepad.cpp
 src/addonmanager.cpp
 src/configmanager.cpp
 src/storagemanager.cpp
+src/system.cpp
 src/configs/webconfig.cpp
 src/addons/analog.cpp
 src/addons/turbo.cpp
--- a/headers/system.h
+++ b/headers/system.h
@@ -1,0 +1,19 @@
+#ifndef SYSTEM_H_
+#define SYSTEM_H_
+
+#include <cstdint>
+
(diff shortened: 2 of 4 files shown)


TheTrainGoes (commented):
This is tested and working on my end.  I got a warning from the address when accessed but that might just be my setup.  

I would like to get a second set of eyes on this before merging.

TheTrainGoes (approved):

arntsonl (approved):
Really cool functionality!

PR #71, opened 2023-01-26
Fix static allocs unit in web-config

deeebug:
This corrects a mistake where static allocations were not handled as bytes.


arntsonl:
Let's rebuild fsdata.c once we get all the PRs in. We'll push in your static allocation changes from WebConfigurator and make a new PR once we're done.

PR #73, opened 2023-01-27
Switch to webconfig mode without a hardware reset

mthiesen:
This changelist allows the user to switch to webconfig mode by pressing and holding `S2` for 4 seconds. Under the hood the hotkey triggers a software reboot using the watchdog of the Pi Pico. The same method is used in MicroPythons implementation of `machine.reset()`. Before reboot a magic value is written to one of the watchdog scratch registers, which is checked in the boot code. Those registers are specifically meant to be used in such a way, they will preserve their contents, even if there is a reboot.

It is also possible to exit webconfig mode again by invoking this endpoint: `http://192.168.7.1/api/reboot`. This should probably be wired to a button on the webconfig at some point.

It is perfectly possible to switch to webconfig mode and back multiple times without the need to pull the USB cable. This should be very useful while tweaking configurations and will probably save the life of many USB ports.

The use of a singular button to trigger this mode is debatable. But I wanted to use the same button that was used before. Accidental triggers should be almost impossible, given the fact that there is a long hold period. Also the hotkey is implemented in such a way that no other buttons are allowed to be pressed at the same time. Random button mashing will do nothing.


mthiesen:
After some discussion on Discord I made the following changes:

- Changed hotkey to `S2 + B3 + B4`, `S2` is an action that is occasionally used in Switch games and could lead to accidental triggers
- Allow existing webconfig with the same hotkey

arntsonl (approved):
Looks great!! The reboot API function call makes this very useful, and I like the use of the scratch registers.

troytroy22:
This helps a lot! Please add the shortcut to readme and documentation.

PR #78, opened 2023-02-01
Make request for setSplashImage atomic

deeebug:
With #77, the API call to set splash image no longer needs to be chunked. This solves potential issues setting the image with certain browsers.


mthiesen (changes requested):
Splash image upload works perfectly for me now, where it would occasionally fail before with my browser.

Just those outdated comments have to be removed.

arntsonl (approved):
Nice, base64 based splash image sounds perfect!

arntsonl (approved):
Nice, base64 based splash image sounds perfect!

PR #81, opened 2023-02-05
Adding delay startup for slow crystals and updates to the build system

Fortinbra:
There are new variables for the build system, so that we can build the fsdata file instead of always committing it to github.


arntsonl (approved):
Looks fantastic! Removing fsdata.c is a game changer. Thanks for all the hard work fort!

arntsonl (approved):

PR #83, opened 2023-02-06
Update Pins to match Pinout Asset

matan157:
Very small update -- I noticed the pins in the config were switched from the asset.

Also some trimmed spaces hehe


arntsonl (approved):
Good catch!

PR #84, opened 2023-02-08
Add-On Enhancements

arntsonl:
Upgrading the functionality of Add-ons to include the following:

- Enable/Disable Add-ons from the Web UI
- Increase total Flash Size to 8kb (from 4kb)
- Separate Add-On Options from Board Options
- Allow for Board-Specified Enable/Disable flags


arntsonl:
Don't merge this.

PR #87, opened 2023-02-09
Changed permissions to make makefsdata executable

thnikk:
CMake currently fails on linux due to makefsdata not being executable.


arntsonl (approved):
Makes sense, eventually it would be nice if we could pull these from their repos instead of including them. But this works!

PR #91, opened 2023-02-10
Fixed dual directional for single pin

arntsonl:
Fixed dual directional so you can assign a single pin (such as just Up) as a dual directional input.


deeebug (commented):
Looks good! 👍

Issue #99, opened 2023-02-14
Add ability to flip display horizontally via boardconfig.h and web-config


TheTrainGoes:
There are a number of new OLEDs hitting AliExpress that appear to have their display flipped.

OneBitDisplay.h supports the `ANGLE_FLIPX` command:

```// Rotation and flip angles to draw tiles
enum {
  ANGLE_0=0,
  ANGLE_90,
  ANGLE_180,
  ANGLE_270,
  ANGLE_FLIPX,
  ANGLE_FLIPY
};```

It would be great to have this added into BoardConfig.H as well as the web-config for anyone using the new displays.

TheTrainGoes:
This was done and merged - thanks @deeebug

PR #100, opened 2023-02-15
Add more display flip modes (for #99)

deeebug:
The following changes address issue: #99

Two new modes have been added with the regular Flip mode.
image

This should help address issues with screens that have their displays flipped by default.


arntsonl (approved):
Looks great!

Issue #101, opened 2023-02-16
Bounty - Ability to map additional instance of a button to a seperate GPIO pin


TheTrainGoes:
Currently you cannot assign multiple instances of a button to different GPIO pins.  

As an example, with UP mapped to GPIO2, I cannot add an additional instance of UP to GPIO22.

This would be an add-on that can be configured in the web-ui.

The first section would let you specify the GPIO that the additional button is hooked up to.
The section section would let you specify which button has the additional instance of assignment.

As an example, you may choose to assign UP to GPIO22, or P1 or even START.

TheTrainGoes:
This has been implemented - thanks @deeebug !

PR #102, opened 2023-02-17
Implement extra button addon

deeebug:
This addon allows an extra button to be mapped to a free pin from the web-config.
Hopefully this is good enough for #101.

image


arntsonl (approved):
Looks great!

PR #103, opened 2023-02-17
Player Number addon

arntsonl:
Added the Player Number X-Input add-on.

This is a useful add-on for an arcade machine that has multiple GP2040-CE devices plugged in.

By setting the player number in X-Input mode, the device will reboot if the incorrect player LED is used by the system. The reboot timing is based on the player number so:

Player 1: No Delay
Player 2: 2s wait before reboot
Player 3: 4s wait before reboot
Player 4: 6s wait before reboot

The lower the player number, the more aggressively it reboots


arntsonl (commented):
Everything works as expected

PR #104, opened 2023-02-18
Player Number Fix

arntsonl:
Disconnect USB while we wait for reboot, this should allow the lower player numbers to connect while the uppers are disconnected


arntsonl (commented):
Works, tested on my arcade

Issue #111, opened 2023-02-20
Add ability to set splash screen length via web-config


TheTrainGoes:
Add the ability to set the length of time that the splash screen is active for via the web-ui.

Options would include:
Default
5 seconds
10 seconds
30 seconds
Always on

arntsonl:
It's in with manual timer

PR #112, opened 2023-02-20
Add splash screen duration feature

deeebug:
This PR is an implementation for feature request in #111.

image

Web-config sets values in ms, except for `Default` which is 0 (customizable in `BoardConfig.h`) and `Always ON` which is -1.


src/storagemanager.cpp
```diff
@@ -93,6 +93,7 @@ void Storage::setDefaultBoardOptions()
 	boardOptions.buttonLayoutRight = BUTTON_LAYOUT_RIGHT;
 	boardOptions.splashMode        = SPLASH_MODE;
 	boardOptions.splashChoice      = SPLASH_CHOICE;
+	boardOptions.splashDuration    = SPLASH_CHOICE;
```
arntsonl, line 96: This should be:

boardOptions.splashDuration    = SPLASH_DURATION;

arntsonl (changes requested):
Add the following to the top of i2cdisplay.h:

#ifndef SPLASH_DURATION
#define SPLASH_DURATION 7500
#endif

this will match our other default defines.

Also need a change where SPLASH_CHOICE needs to be set to SPLASH_DURATION

arntsonl (approved):
Looks good!

PR #113, opened 2023-02-21
update with community splash images

TheTrainGoes:


arntsonl (approved):
Looks awesome!! This will be a great addition for folks wanting custom splash screens

PR #117, opened 2023-02-26
Fix default splash duration value

deeebug:
This change fixes the Display Configuration page's default state which is flagged as invalid.


arntsonl (changes requested):
We should not be doing logic for default assignment outside of storage manager. We need to revisit this and refactor some code.

Issue #120, opened 2023-02-28
Splash Display Fix + Code Clean-up


arntsonl:
Splash display is not respecting the default time properly

We need to look at the logic for #define SPLASH_DISPLAY

Additionally, we started having Get* functions in the StorageManager.h which breaks our current coding conventions. We need to remove these and make the StorageManager.h get functions single option only rather than option.<type>

This means we need to get rid of 

int Storage::GetButtonLayout()}
int Storage::GetButtonLayoutRight()
int Storage::GetButtonLayout()
int Storage::GetButtonLayoutRight()
int Storage::GetSplashMode()
int Storage::GetSplashChoice()
int Storage::GetSplashDuration()
int Storage::GetSplashMode()
int Storage::GetSplashChoice()
int Storage::GetSplashDuration()

and start being more strict about PRs that do this.

TheTrainGoes:
This has been addressed.  Thank you @deeebug and @arntsonl

PR #121, opened 2023-03-01
Refactor and cleanup storagemanager and i2cdisplay (#120)

deeebug:
I've attempted to address some problems with the said code and refactor accordingly.
With these changes, I'll also fix the default splash duration as per request soon.

Please review these changes, thanks!


arntsonl (dismissed):
Muuuuuch cleaner, this is good clean code. I changed one double assignment for readability, the assembly code is the same so you don't save anything.

PR #129, opened 2023-03-30
Bounty Shmup Mode - Extended Turbo

arntsonl:
Bounty for Shmup Mode

This extends the functionality of Turbo to include the following:

- 4 Always-on Turbo Button Masks
- 4 Charge Shot Pins w/ Masks
- Turbo Dial Support (ADC pin)
- Turbo or Charge Priority Modes (charge + turbo pressed at the same time)


arntsonl (commented):
Should be good

Issue #133, opened 2023-04-08
Bounty - SOCD slider switch


TheTrainGoes:
Like the currently implemented LS/RS/DP slider we are hosting a bounty to have an SOCD slider switch added in to the addon section.

Conditions for completion of the bounty:
- Added into the code in the same way our other addons are done
- Tested and working
- Does not impact other addons
- Added into web-config

The default state of this addon should be off.  It should use two GPIO pins.  The default state should be SOCD neutral as it is now.

arntsonl:
It's in release! Closing this, nice work @deeebug !

PR #134, opened 2023-04-08
Implement SOCD Slider addon

### LINKED ISSUE #132: Add-on request - SOCD slider toggle
Piggybacking on the existing add-on for the LS/DP/RS add-on, there has been a request for the same thing to have a dedicated SOCD slider toggle.  

This has become an increasingly requested add-on with the new SOCD rules which would allow people to quickly change as they learn new play styles.

deeebug:
This adds a feature that allows changing SOCD with slider input.
It also should close #132 and #133.

image


arntsonl (approved):
Looks good!

PR #138, opened 2023-04-10
Removing MPG from GP2040-CE

arntsonl:
This cleans the MPG instance from GP2040-CE, using the same code as MPG had but removing the extra layer.

We still need to clean this code as its pretty messy and has WAY too many inlines and statics, but this is the first step!


arntsonl (commented):
Checked everything over, looks good to me

PR #150, opened 2023-04-16
Implement Daughter Board Add-on

Caian:
This is the implementation of a Daughter Board Add-on (and simple communication protocol) that can be used by any microcontroller to extend the number of analog and digital inputs of the main GP2040 board.

The data protocol is defined as follows:

- Each data response from the daughter board has 16 words.
- Each word is a 2-byte little-endian unsigned integer.
- The first words are the analog values, and the rest are the digital values.
- Each analog value uses exactly one word (0-65535).
- Each digital value uses one bit of a word.
- Each word can fit 16 digital values.

The configuration of the daughter board is not covered by the protocol and it is done separately.

### Input Mapping

The input is mapped using the `Daughter Board Analog Map` and `Daughter Board Digital Map` fields of the Daughter Board Add-on. These are comma-separated values of the MPG inputs that will be associated with each word/bit of the response data, in the order as they appear. Analog inputs are handled first, then the digital inputs.

A reference implementation using ESP32 and PlatformIO is here: https://github.com/Caian/gp2040-esp32-daughter-board

**This feature requires my feature request on the WebConfigurator repo**


headers/storagemanager.h
```diff
@@ -100,6 +101,13 @@ struct AddonOptions {
 	int i2cAnalog1219Block;
 	uint32_t i2cAnalog1219Speed;
 	uint8_t i2cAnalog1219Address;
+	uint8_t gpDaughterSDAPin;
```
FeralAI, line 104: Options that will persist to EEPROM need to stay in order to avoid breaking saved configs on previous versions. These new fields should go before the `checksum` field.

Caian: Done

arntsonl:
I've spoken with @Caian and requested a proof of concept using a Pico for the daughter board as well as some other naming conventions and code changes.

Once we have a daughter board that everyone can test with, this will allow all developers on the project to maintain support for the add-on and test various changes.

We also should look at making the I2C communication protocol some kind of standard either based on the designs of another project or implementing our own structure. This will give us the best flexibility and maintainability in the future.

arntsonl:
Closing this one until we have a working Pico implementation of a daughter board. We want to make sure everyone can test these things when we do add-on changes, sometimes that means sending a coder a piece of hardware or setting something up.

RobertDaleSmith:
Hey @Caian, great work with getting this effort started!

How tough would it be to make this add-on work in both directions? My idea before finding this PR was to create an add-on to allow multiple GP2040-CE boards to communicate bi-directionally. With both sides sending and receiving each other's buttons. Both being capable of being the daughter board depending on which one is connected to USB.

Using two Adafruit KB2040's, one would be able to easily connect the two together with a simple STEMMA cable and they could share each others inputs and power. Not sure if both sending buttons would just be unnecessary latency or if that diff is negligible. Either way I think at the very least we need an I2C mode that the desired daughter pico board could run. It is excellent that one could use any other board but I think having a GP2040-CE based daughter would be ideal for maintaining the feature long term. 

I pulled this branch to start poking around but got hung up with merge conflicts. If you are able to resolve those and/or add I2C writing capability to either the add-on or as a boot mode, then I'm more than willing to help test and push this over the finish line.

PR #152, opened 2023-04-16
Fixed pin checks which also fixes our input latency

arntsonl:
Fixed pin checks for Reverse and Turbo, as well as a Web Config pin c…heck that was missed.

Also decreased minimum Turbo shots from 5 to 2 (much slower).


arntsonl (commented):
Works! This fixes our issue, we're good for 0.6.3

PR #161, opened 2023-04-20
Add keyboard mode implementation

deeebug:
As the title describes, this adds a new input mode.
The mappings are configurable and the mode can be switched to during boot with the B4 button, or via the web-config, as other input modes.

For key mapping, there's a new key mapping page similar to pin mapping.


mthiesen (approved):
Looks great! Can be merged as is as far as I am concerned.

arntsonl (approved):
Super clean, thanks for implementing this @deeebug !!

PR #164, opened 2023-04-21
Disable turbo by default

TheTrainGoes:
This PR disables Turbo by default in the following configs:

Pico
PicoAnn
SparkFunProMicro
WaveshareZero


arntsonl (approved):
Looks good, makes sense for our default config

Issue #107, opened 2023-02-19
Bounty - New LED mode - Individually set colors for on/off state via web-config


TheTrainGoes:
After some conversation in the BentoBox channel I am throwing a bounty up to see if we can get a new LED mode added that allows someone to configure a static on/off color per button / LED via web-config.

As an example, if I have a stickless layout I could make LEFT, DOWN, RIGHT, UP green and the rest red.  Then on press LEFT, DOWN, RIGHT, UP could turn blue while the rest could turn pink.

We already define colors in the code so those could be whats provided via the web-config.  

The web-config would have a section in the LED configuration to deal with this new mode.

FeralAI:
I'm starting work on this feature. Currently targeting the minimum viable implementation at the moment since there's a bigger LED refactor coming along.

TheTrainGoes:
This has been merged

PR #166, opened 2023-04-22
Custom LED implementation

FeralAI:
This is the firmware implementation for https://github.com/OpenStickCommunity/GP2040-CE/issues/107.

This PR requires https://github.com/OpenStickCommunity/WebConfigurator/pull/38 for the UI updates.

A new animation mode for Custom Theme has been added. If enabled, this mode will use the new CustomTheme and CustomThemePressed animation subclasses and configured LED options to display a customized static theme with customized button color presses, as the issues requested.

In the process of debugging I reworked how the pressed animation is created since it was only created as a fallback instead of when needed like the base animations.

If we want to do any customized themes with animations that will require a larger refactor or migration of the LED controller library.


FeralAI:
This should be functionally ready now. Summary of changes not directly related to this feature:

* Fixed excessive memory use in `webconfig.cpp` when deserializing POST data
* Bumped stack size up to 4kb with `add_compile_definitions(PICO_STACK_SIZE=0x1000)` in `CMakeLists.txt`
* Updated ArduinoJson library to v6.21.2
* Fixed issue where user couldn't use the `Previous Animation` shortcut to cycle around the animation list (needed explicit cast to check for negative)
* Pressed animations are now properly created and destroyed like normal animations
* Pressed animations should flicker much less when changing animations

arntsonl (approved):
Super cool!! With fluppi's changes, we'll have better stack space and fully independent LED configurations!

PR #167, opened 2023-04-22
Switch extra button Dpad mask to defines instead of magic numbers

deeebug:
This does away with magic numbers for Dpad related defines in BoardConfig.h.
This also fixes a bug that would have occurred with Right Dpad button.

The defines are located in GamepadState.h and are:

GAMEPAD_MASK_DU
GAMEPAD_MASK_DD
GAMEPAD_MASK_DL
GAMEPAD_MASK_DR


arntsonl (approved):
Makes good sense to me!

PR #169, opened 2023-04-22
Update development docs to include NodeJS

n-i-x:
GP2040-CE cannot be built by following the instructions if NodeJS is not already installed. Added instructions for installing NodeJS.


TheTrainGoes:
@deeebug if you can check this out and approve if accurate that would be great.

TheTrainGoes (approved):

PR #172, opened 2023-04-23
Fixed bug where webconfig could display stale data

mthiesen:
Previously `/api/getButtonMapping` did not reflect changes made to the button mapping during the current webconfig session.
The same was true for the `usedPins` array in `api/getLedOptions`, `api/getAddonOptions` and `api/getDisplayOptions.

This PR fixes all these issues.


deeebug (approved):
Tested and working fine!

arntsonl (approved):
Yeah we gotta do something about the uint8 business, but this is a good fix for now. Thanks!

PR #174, opened 2023-04-23
Handle case where invalid pin is configured 

mthiesen:
Previously the case where a gamepad pin was unassigned was not handled correctly. We would shift by (uint8_t)(-1) which is undefined behavior in C++.


arntsonl (approved):
Nice, these are good. We definitely need some revamping on our pin assignment functions.

Issue #160, opened 2023-04-18
Remove Frame Sleep() as Pico does a busy wait


arntsonl:
Welp.... this was my bad. Thanks @mthiesen for pointing out that sleep() on the Pico does a busy wait and does not give back anything. Deep sleep is the only way to reduce power usage according to the Raspberry Pi Pico SDK.

We can remove sleep and just rely on our timers to make sure we hit polling as fast as possible.

arntsonl:
Turns out the sleep() doesn't impact lag, and actually does return a bit of cpu back to reduce power.

PR #175, opened 2023-04-23
Remove sleeps

mthiesen:
This fixes issue #160.

My previous assumption that `sleep_us()` is implemented with a busy loop turned out to be incorrect. In fact the Pi Pico SDK tries to save power by using the WFE instruction. While the power savings are measureable they are minimal. I have measured a power consumpsion of 21.5 mA without `sleep_us()` vs. 20 mA with `sleep_us()`.

Whether this PR should be merged should depend on its effect on input latency. Unfortunately, I currently don't have the ability to do latency measurements.


arntsonl:
Before we approve/merge, we need to visit if removing the waits does anything positive/negative to our GPIO and I2C rates. I'll test on my input latency tester and have others test on theirs (if anyone has a Mister, that would be awesome)

arntsonl:
Here is our current 0.7.0 Pre-Release with sleep, delay, and no add-ons enabled:

`polling interval    1.0
min    0.54
max    1.34
average    0.83
stddev    0.24
% no skip    96.05%
% skip 1 frame    3.95%
% skip 2 frames    0%
% skip 3 frames    0%`

Here is our current 0.7.0 Pre-Release with no sleep, delay, and no add-ons enabled:

`polling interval    1.0
min    0.54
max    1.38
average    0.85
stddev    0.24
% no skip    95.95%
% skip 1 frame    4.05%
% skip 2 frames    0%
% skip 3 frames    0%`

Here's 0.6.3:

`polling interval    1.0
min    0.54
max    1.38
average    0.86
stddev    0.24
% no skip    95.91%
% skip 1 frame    4.09%
% skip 2 frames    0%
% skip 3 frames    0%`

and 0.4.3:

`polling interval    1.0
min    0.54
max    1.39
average    0.87
stddev    0.24
% no skip    95.79%
% skip 1 frame    4.21%
% skip 2 frames    0%
% skip 3 frames    0%`

removing the delay causes errors/input drops on the input latency tester and most likely calls tinyusb too quickly. I'm closing this one, we can revisit if there's any way to increase USB polling on the clock-rise but we can't remove delays unfortunately.

mthiesen:
@arntsonl:

Thank you for checking this out. I'll look into latency myself again once I have a working latency test setup.

You should probably also close #160 as well.

PR #177, opened 2023-04-25
Add a way to get defaults for input and D-Pad mode

deeebug:
This adds a method to define default parameters for Input and D-Pad mode in `BoardConfig.h`, which is optional.
Please accept if we want to support this.


smashBrosKanu (approved):
Thanks!

smashBrosKanu (approved):
Thanks!

smashBrosKanu (approved):

arntsonl (approved):
Good feature!

PR #179, opened 2023-04-26
assign A2 button to a button event in HID mode

bsstephan:
we declare 14 buttons (all active in Switch mode), but D-Input mode only declares 13 input descriptors and ignores the A2 event. this button won't do anything on a PS3, but Linux (or MiSTer, etc.) could use this button for something.

"tp" name chosen by PlayStation convention, since on a DS4 the touchpad registers as button 14.

### DIFF
--- a/headers/gamepad/descriptors/HIDDescriptors.h
+++ b/headers/gamepad/descriptors/HIDDescriptors.h
@@ -83,6 +83,7 @@ typedef struct __attribute((packed, aligned(1)))
 	uint8_t r3_btn : 1;
 	
 	uint8_t ps_btn : 1;
+	uint8_t tp_btn : 1;
 //	uint8_t l2_btn_alt : 1;
 	
 //	uint8_t r2_btn_alt : 1;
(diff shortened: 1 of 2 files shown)


TheTrainGoes:
Hey @bsstephan ,

We need to test this out on the PS3 and make sure everything still works as intended.  Part of the trouble we ran into with DINPUT was getting the home button to work on the PS3.  

Should this change prevent that from working we may look at splitting DINPUT into two separate modes - PS3 mode and generic DINPUT which would allow for both the proper button mappings for PS3 and also the 14 inputs for DINPUT.

bsstephan:
I tried my build on PS3 in GGXXAC+R and all the buttons mapped as expected in the controller config. Home button also worked to register the controller with the PS3, and in game, home button also worked as expected (went to XMB, and pressing again went back to game). More testing is obviously a good idea, but it passed my sniff test.

@TheTrainGoes I will do a build for base Pico now and send it your way.

arntsonl (approved):
Looks good, thanks for the add!

PR #183, opened 2023-04-28
Add SOCD First Input Priority to hotkeys list

deeebug:
With the addition of the First Input Priority SOCD type, the hotkeys configuration list also needs to be updated. This PR addresses that.

Please review and merge, thank you!


arntsonl (approved):
Looks good, simple change

PR #189, opened 2023-04-30
Super simple Analog fix

arntsonl:
Stupid bug on my part, swapped analog pins. Also made it so you can run analog without setting the pins to match our other add-ons.


arntsonl (commented):
Really dumb easy fix

PR #200, opened 2023-05-01
Fix unwanted cleaning of F1 and F2 button masks

deeebug:
This fixes an issue that unnecessarily cleans F1 and F2 button masks even when a valid hotkey combo isn't pressed.
Apologies for the problem.

Please check and merge, thank you!


arntsonl (approved):
Looks good, only apply cleaning on hotfix action

PR #205, opened 2023-05-02
Storage manager optimizations

mthiesen:
Currently `StorageManager` returns its configs by value, which is unnecessary and inefficient. In particular this is a problem with the Turbo addon, which reads its config on every `process()` call.

With this PR `StorageManager` returns configs as a `const` reference. I have also moved the getters to the header so that they can be inlined by the compiler.


arntsonl (approved):
Nice thank you for taking care of this!

This has been on my to-do list for a while. I had to use & for ps4Options because of memory limitations, and this takes care of everything else!

PR #208, opened 2023-05-03
Send only changed values in Addons page to the board

deeebug:
For any change, big or small, the addons config page sends all of the data to the board, which is inefficient and also caused a recent bug. This PR aims to send only what's updated and the firmware also accounts for fields which have been omitted.

Web-config repo has also been updated for this (https://github.com/OpenStickCommunity/WebConfigurator/pull/55).


arntsonl (approved):
So fresh, so clean!

PR #219, opened 2023-05-06
Updating CMAKE with new public action

Fortinbra:
I merged a few steps together, and created a new action from them. Published said new action publicly, so anyone can add it to their pipelines. Now we're using it too!


arntsonl (approved):
Looks clean!

PR #222, opened 2023-05-07
Add Reflex Board to the configs

TheTrainGoes:
This adds the Reflex board from MiSTer Addons to the config section.


arntsonl (approved):
Nice! We can talk to the group about adding this to our uf2 makes, we also have a neat way of doing that coming up.

PR #232, opened 2023-05-09
Escape release notes headers from showing up

deeebug:
Release notes from the Github releases API were causing the notes sections to be displayed in sidebar, which may be unwanted and are causing layout inconsistencies.

In the following image, the sub-sections are not part of the `Installation` page.

image


FeralAI (approved):
👍

PR #233, opened 2023-05-09
Disable Wii extension addon by default

FeralAI:
The Wii extension addon inadvertently enables the addon and maps pins by default, causing validation issues in the web config and confusion for the user. This PR disables the extension by default and unsets the default pin mappings.

### DIFF
--- a/headers/addons/wiiext.h
+++ b/headers/addons/wiiext.h
@@ -14,7 +14,7 @@
 #define WiiExtensionName "WiiExtension"
 
 #ifndef WII_EXTENSION_ENABLED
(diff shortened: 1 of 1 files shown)


mthiesen (approved):

bsstephan:
FYI `GP2040-CE_0.7.1_MavercadeKeebfighter.uf2` does not boot into any controller without this change. `git bisect` narrowed the boot issue down to d662fcbdfea1bd4ff229224e0497ec90bb786bae and merging this lets it boot again; must be some pin funny business.

PR #234, opened 2023-05-09
use getSOCDMode in DDI None to benefit from 28d47d9f

bsstephan:
Just a tiny code cleanup/improvement to my previous PR to account for the no SOCD mode change that also went in.

### DIFF
--- a/src/addons/dualdirectional.cpp
+++ b/src/addons/dualdirectional.cpp
@@ -157,7 +157,13 @@ void DualDirectionalInput::process()
             // this also avoids accidentally masking gamepad inputs with the lack of dual inputs
             if (gamepad->options.dpadMode == options.dualDirDpadMode) {
                 uint8_t gamepadDpad = gpadToBinary(gamepad->options.dpadMode, gamepad->state);
-                dualOut = SOCDGamepadClean(dualOut | gamepadDpad, socdMode == SOCD_MODE_SECOND_INPUT_PRIORITY);
+                if ( socdMode == SOCD_MODE_NEUTRAL ) {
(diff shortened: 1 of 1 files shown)


src/addons/dualdirectional.cpp
```diff
@@ -157,7 +157,7 @@ void DualDirectionalInput::process()
             // this also avoids accidentally masking gamepad inputs with the lack of dual inputs
             if (gamepad->options.dpadMode == options.dualDirDpadMode) {
```
deeebug, line 158: I've done some tests and this seems to be the logic that gets both regular and DDI working together when both input methods are same.

```c
if (gamepad->options.dpadMode == options.dualDirDpadMode) {
    uint8_t gamepadDpad = gpadToBinary(gamepad->options.dpadMode, gamepad->state);
    if ( socdMode == SOCD_MODE_NEUTRAL ) {
        dualOut = SOCDCombine(socdMode, gamepadDpad);
    } else if ( socdMode != SOCD_MODE_BYPASS ) {
        dualOut = SOCDGamepadClean(dualOut | gamepadDpad, socdMode == SOCD_MODE_SECOND_INPUT_PRIORITY);
    } else {
        dualOut |= gamepadDpad;
    }
}```

I think line 158 to 161 will be replaced by this change.

deeebug:
Thank you for the update, looking good to me!
I've done some tests and suggested a bit of a change, so please look into it.

bsstephan:
@deeebug I believe this also makes sense to me!

arntsonl (approved):
Nice, SOCD with dual is so confusing but glad we're getting some solid fixes!

PR #241, opened 2023-05-11
Bug fixes and tweaks for Wii extension support

mikepparks:
Added improved Wii extension detection.
Added GHWT touch fret support.
Added initial Wii extension hot plugging support.
Fixed issue with detecting GHWT guitars.
Fixed issue with guitar analog values.
Mapped taiko controller left drumpad strikes to D-Pad Left instead of Right


arntsonl (approved):
Keeps getting better and better!

PR #248, opened 2023-05-14
Adds kb2040 board config

RobertDaleSmith:
Adds base KB2040 board config used in my recent Fisher-Price controller mod.


FeralAI (commented):
Looks good. The pin mapping diagram should also be added to the Wiring section of the doc site.

RobertDaleSmith:
> Looks good. The pin mapping diagram should also be added to the Wiring section of the doc site.

Cool thanks for taking a look. Is the doc site a public repo I can push a PR to?

FeralAI:
> Cool thanks for taking a look. Is the doc site a public repo I can push a PR to?

The docs site is embedded in this repo in the `docs` folder. You can copy the pin diagram to the `docs/assets/wiring` folder then update `docs/wiring.md` with the new section.

configs/KB2040/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+#add_executable(KB2040
```
mthiesen, line 1: This file is not necessary. I know that there is such a file in the `configs/Pico` directory but that one is probably unnecessary as well.

mthiesen (changes requested):
Apart from the extra file everything looks fine to me.

RobertDaleSmith:
@FeralAI pin mapping diagram for KB2040 has now been added to docs.

FeralAI (approved):
Thanks for updating the docs!

www
deeebug, line 1: Might want to exclude this one from the changes.
This submodule is handled separately.

deeebug (changes requested):
Everything looks great to me!
Just that the `www` submodule needs to be excluded from the changes.

arntsonl (approved):
Looks great!

PR #250, opened 2023-05-15
Www merge

Fortinbra:
Merging webconfig back into main build


mthiesen (approved):

deeebug:
Noticing this file here:
https://github.com/OpenStickCommunity/GP2040-CE/blob/043b53a960d01fe817c8d5aac886e6131e7e37e8/www/lib/httpd/fsdata.c
Not sure if this should be here, probably can be ignored and cleaned up later rather than complicating this one.

Other than that good with everything!

PR #251, opened 2023-05-15
Fix PLED defines being used instead of new LEDOptions

FeralAI:
Bug fix from PR https://github.com/OpenStickCommunity/GP2040-CE/pull/246


arntsonl (approved):
Show stopper bug for RC2

PR #253, opened 2023-05-15
Fix PLED_TYPE overriding LEDOptions.pledType

FeralAI:
Missed this update when implementing the PLED configuration. Firmware was always using the PLED_TYPE instead of the configured option.


arntsonl (approved):
Makes sense, move things to storagemanager.

PR #263, opened 2023-05-19
Protobuf Config

mthiesen:
This is the promised Protobuf refactoring. This PR contains A LOT of changes:

- Configuration is now stored with Protobuf serialization using the [Nanopb](https://github.com/nanopb/nanopb) library. This will allow us to change our configuration between versions (within limits) without invalidating the configs stored by users. The two Protobuf source files are `proto/config.proto` and `proto/enums.proto`.
- Protobuf compilation is automated with CMake. Compilation is only invoked when the `.proto` file actually changed. The only requirement is a Python3 installation.
- Config conversion from / to JSON. This is implemented using the reflection information generated by Nanopb, there is no need for manual adjustment when the Protobuf schema changes. This does not support all Protobuf features right now, but covers what we need currently and more. Can be extended as needed.
- `api/getConfig` and `api/setConfig` endpoints are implemented in Webconfig to read / write the whole config at once. They use the JSON functionality and will also adjust to schema changes. `api/setConfig` will correctly return 400 if it gets passed invalid JSON and 500 if saving fails for whatever reason (it should not).
- I took the chance to change the structure of config. In particular I split `BoardConfig` into multiple `struct`s as well as `AddonConfig`. I used the structure of Webconfig UI and the Webconfig API as a rough guideline.
- Migration from 0.7.1. On first boot the legacy config will be converted to Protobuf config. If everything goes well users should not notice any change and will not lose any settings.

The following was tested by me:
- Loading / saving
- Building with GitHub Actions
- Migration from 0.7.1 to Protobuf
- Booting after flash nuke
- Webconfig API
- Changing Input Mode / D-Pad Mode / SOCD Cleaning Mode
- Changing Hotkey Settings
- Changing Pin Mapping
- Changing Keyboard Mapping
- Changing Display Configuration
- Input Reverse Addon
- On-Board LED Addon
- BOOTSEL Button Config (disables my OLED, but the Webconfig UI says this can happen)

The following addons should still be tested by those with the required hardware:
- Analog Addon
- Turbo Addon
- Joystick Selection Slider Addon
- I2C Analog ADS1219 Addon
- Dual Directional Input Addon
- Buzzer Speaker Addon
- Extra Button Addon
- Player Number Addon
- SOCD Cleaning Mode Selection Slider Addon
- PS4 Mode Addon
- Wii Extension Configuration Addon
- LED Configuration
- Custom LED Theme

I'd also like to hear feedback for the current config schema. If we want to change something it will be much easier to do this now than later.

I will provide proper documentation that explains how to change the config schema with an additional PR.


henrebotha:
> Configuration is not stored with Protobuf serialization

Typo?

mthiesen:
> > Configuration is not stored with Protobuf serialization
> 
> Typo?

Yes, fixed. Thanks.

src/addons/jslider.cpp
```diff
@@ -60,8 +62,8 @@ void JSliderInput::process()
 #endif
 
     Gamepad * gamepad = Storage::getInstance().GetGamepad();
-    if ( gamepad->options.dpadMode != dpadState) {
-        gamepad->options.dpadMode = dpadState;
+    if ( gamepad->getOptions().dpadMode != dpadState) {
+        gamepad->getOptions().dpadMode == dpadState;
```
bsstephan, line 66: LD/DP/RS selection doesn't seem to do anything anymore, I think line 66 here is errant and the cause.

bsstephan: This fixes it:

```
diff --git a/headers/gamepad.h b/headers/gamepad.h
index 0328c9c..bbaf85f 100644
--- a/headers/gamepad.h
+++ b/headers/gamepad.h
@@ -126,6 +126,7 @@ public:
 
 	const GamepadOptions& getOptions() const { return options; }
 
+	void setDpadMode(DpadMode dpadMode) { options.dpadMode = dpadMode; }
 	void setInputMode(InputMode inputMode) { options.inputMode = inputMode; }
 	void setSOCDMode(SOCDMode socdMode) { options.socdMode = socdMode; }
 
diff --git a/src/addons/jslider.cpp b/src/addons/jslider.cpp
index 04ab776..d30302e 100644
--- a/src/addons/jslider.cpp
+++ b/src/addons/jslider.cpp
@@ -63,7 +63,7 @@ void JSliderInput::process()
 
     Gamepad * gamepad = Storage::getInstance().GetGamepad();
     if ( gamepad->getOptions().dpadMode != dpadState) {
-        gamepad->getOptions().dpadMode == dpadState;
+        gamepad->setDpadMode(dpadState);
         gamepad->save();
     }
 }
```

mthiesen: @bsstephan I just submitted a near identical fix :)

Good catch!

src/config_legacy.cpp
```diff
@@ -0,0 +1,1090 @@
...
+        int i2cBlock;
+        uint32_t i2cSpeed;
+        bool hasI2CDisplay;
+        int displayI2CAddress;
+        uint8_t displaySize;
+        uint8_t displayFlip;
+        bool displayInvert;
+        int displaySaverTimeout;
+        ButtonLayoutCustomOptions buttonLayoutCustomOptions;
+        char boardVersion[32]; // 32-char limit to board name
```
arntsonl, line 430: We might be able to use boardVersion when doing legacy conversions

arntsonl (dismissed):
Code looks good! I didn't checkbox the 500+ protobuf library files, but I looked at everything else. Lots of various fixes and improvements, this is a big change so we'll have to do some merging with all the PR features coming in.

arntsonl:
Bug #1: Custom LED enable mode does not save

1. Go to web configurator
2. Click enable and create a custom LED theme
3. Click save, and restart back into web configurator
4. Navigate back to Custom LED theme and observe the option is no longer selected

arntsonl (approved):
Fixed!

PR #264, opened 2023-05-19
Update references to GP2040 in www

henrebotha:


arntsonl (approved):
Nice, good clean-up thank you.

PR #266, opened 2023-05-19
Update button labels for button label change and TP-Switch swap

deeebug:
Previous PR (#254) had an issue, where in certain cases, the web-config would complain about invalid boolean data. This fixes that. In addition, the button label choice and button swap for PS4 is also reflected in various parts of the web-config and is synced across browser tabs.

An older constants file for Buttons has also been cleaned up.

Please verify the changes and merge, thank you!


arntsonl (approved):
Looks like a good clean-up!

PR #267, opened 2023-05-21
.gitignore a couple files used by vim + tools

bsstephan:
These are a couple temporary files and directories that are made by using vim + YCM + clangd as an IDE.


arntsonl (approved):
Makes sense, keeping it clean!

Issue #272, opened 2023-05-23
Bounty - Forced setup mode


TheTrainGoes:
Forced setup mode is intended for operators and individuals who may be putting these boards into dedicated setups and do not want the ability to change options once installed.

This bounty would be either an add-on or an addition to the settings page.

It will allow you to:
- Lock a mode (holding P1, P2, K1, K2 on boot no longer changes modes from the set mode)
- Remove any or all hotkey combos (this is already in, not sure if we want to wrap it all together)

bsstephan:
It might be worth a quick discussion at some point if the "Function" key (#239) ought to solve this problem by essentially claiming mode selection, web configurator at boot, etc. are hotkeys, and then the setup mode requested here would be enabled by defining the pin for Fn (which is unset by default, but could be set to a pin only populated via a jumper or on a button/switch not accessible by users). Function (currently a core button) could change --- it's in a WIP waiting for at least protobuf to merge, right now, anyway --- so we have time to redesign it before it goes in.

And to be fair, it's equally possible that "Function" serves as a runtime alternate function key (modifiers e.g. tilts, as BRP said on Discord), but what I was envisioning was more of a "disabling hot keys" thing than a "changing behavior" thing (though both are valuable). If we want "Function" to be runtime then I may un/redo #239 and take this on separately.

Or maybe "forced setup mode" is different than "disabling hot keys" in which case we have three buttons?? I don't know. Anyone with opinions, ping me. =)

TheTrainGoes:
This work has been completed.

PR #273, opened 2023-05-23
Implement Forced Setup Mode & Hotkey Lock settings

deeebug:
This PR adds a drop down for Forced Setup Mode with options:
1. Off
2. Disable Input-Mode Switch
3. Disable Web-Config
4. Disable both Input-Mode Switch and Web-Config

With that it also adds a switch to disable hotkeys to the bottom of the hotkey configuration menu.
These options function as per the requirement outlined in #272.

An alternative suggestion has been made to achieve this function by @bsstephan (https://github.com/OpenStickCommunity/GP2040-CE/issues/272#issuecomment-1559469175)

Please add any comments or suggestions regarding this.
Thanks!


TheTrainGoes:
@deeebug - quick question for you

When either of these modes are enabled:
3 - Disable Web-Config
4 - Disable both Input-Mode Switch and Web-Config

How do you get back into web-conifg?  Or do you need to nuke at that point?

deeebug:
@TheTrainGoes needs a nuke for now, maybe the solution would be to make it be harder to set?

deeebug:
Just added a warning prompt with a basic check to prevent unintended saving of this setting.

mthiesen (approved):

PR #279, opened 2023-05-24
Fix invalid reference to button label

deeebug:
This fixes an issue introduced in #266. It turns out the Custom LED page was left out from the change.


arntsonl (approved):
Good fix!

PR #287, opened 2023-05-27
Implement Keyboard Host via PIO

deeebug:
The following changes integrates [Pico-PIO-USB](https://github.com/sekigon-gonnoc/Pico-PIO-USB) to implement a keyboard host interface. With this, a keyboard can be connected via GPIO pins to have its inputs translated via GP2040-CE.

image

Care must be taken when wiring as incorrectly mapping the GND could cause the keyboard to be borked (happened to one).
Currently, D+ and D- pins are set to GPIO0 and GPIO1 pins. RGB LED also seems to be incompatible with this addon.

This is only one implementation and needs to be expanded to support NKRO, which will be done soon. In the future, several other USB devices could be supported. I think the Pico-PIO-USB integration can be very useful for many use-cases, for instance, it could also be used as an adapter that reserves player slot, given that host support for other controllers are implemented.

Please review the changes and merge if all is good,
Thanks!


src/addons/keyboard_host.cpp
```diff
@@ -0,0 +1,197 @@
...
+KeyboardButtonMapping *_keyboard_host_mapButtonR3;
+KeyboardButtonMapping *_keyboard_host_mapButtonA1;
+KeyboardButtonMapping *_keyboard_host_mapButtonA2;
+
+bool KeyboardHostAddon::available() {
+	return Storage::getInstance().getAddonOptions().keyboardHostOptions.enabled;
+}
+
+void KeyboardHostAddon::setup() {
+  KeyboardMapping& keyboardMapping = Storage::getInstance().getAddonOptions().keyboardHostOptions.mapping;
```
mthiesen, line 32: This can be a `const` reference. You are not modifying the `struct` behind the reference.

src/addons/keyboard_host.cpp
```diff
@@ -0,0 +1,197 @@
...
+  _keyboard_host_state.ry = GAMEPAD_JOYSTICK_MID; 
+  _keyboard_host_state.lt = 0;
+  _keyboard_host_state.rt = 0;
+
+  for(uint8_t i=0; i<7; i++)
+  {
+    uint8_t keycode = (i < 6) ? report->keycode[i] : getKeycodeFromModifier(report->modifier);
+    if ( keycode )
+    {
+      GamepadOptions& gamepadOptions = Storage::getInstance().getGamepadOptions();
```
mthiesen, line 141: This can be a `const` reference. You are not modifying the `struct` behind the reference.

mthiesen (changes requested):

deeebug:
@mthiesen thank you very much for the review! 
I'll get these changes added now.

mthiesen (approved):

ez2torta:
This PR looks so awesome! I don't have access to a rpi pico until one more month but I'm sure I will test this ASAP 💕💕

www/src/Pages/AddonsConfigPage.jsx
```diff
@@ -617,9 +624,40 @@ function flattenObject(object) {
...
+
+	const handleKeyChange = (values, setFieldValue) => (value, button) => {
+		const newMappings = {...values.keyboardHostMap};
+		newMappings[button].key = value;
+		const mappings = validateMappings(newMappings);
+		setFieldValue('keyboardHostMap', mappings);
+		setValidated(true);
+	};
+
+	// const handleSubmit = async (e) => {
```
arntsonl, line 640: We can delete any big comment blocks :)

arntsonl (commented):
Only other change is adding D+/D- to our Web Config (if we can, if not lets me just add this) and we're ready to merge!

arntsonl (approved):
Looks great

www/server/app.js
```diff
@@ -175,7 +174,7 @@ app.get("/api/getPinMappings", (req, res) => {
 });
 
 app.get("/api/getKeyMappings", (req, res) =>
-	res.send(mapValues(keyboardMapping))
+	res.send(mapValues(DEFAULT_KEYBOARD_MAPPING))
```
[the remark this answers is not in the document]
deeebug: This is just the default data sent to the dev server, so it's good.

Issue #243, opened 2023-05-12
Dark-Mode for Web-Config


lost-in-spacetime:
Playing around last night in the web-config made me realize how bright the page can be when full screen at night.

I'm currently using the plugin Dark Reader to remedy the issue for myself. 

It was suggested that I create an issue, possibly an easy update to add Dark-Mode to the config? 

Thanks for entertaining my idea!

TheTrainGoes:
This sounds like a good bounty candidate.  Dark mode slider?

FeralAI:
Very nice @SavageCore! I'm ok bringing in the Bootstrap v5.3 alpha dependecy for now. Our use of Boostrap is quite limited, so I doubt rolling with this version would be an issue until the stable version is released.

Just a few things and I think this could be a PR:

* Could `DarkMode.js` be renamed to `ColorScheme.js`? That seems more fitting since this could in theory be used for a variety of themes.
* We've started to consolidate localStorage options in `AppContext.js` and the `theme` setting would go there

PR #295, opened 2023-05-29
Feature/webui dark mode

SavageCore:
Addressing @FeralAI's comments in [#243](https://github.com/OpenStickCommunity/GP2040-CE/issues/243#issuecomment-1566280103)

Let me know if you want the Dropdown (DANGER MODE) button changed in a separate PR


Pelsin:
Tested, works and looks really nice!

FeralAI:
The code looks great, just a few issues with the move to Bootstrap v5.3.

* The navbar background is now a lighter gray. Maybe not too bad, but there's less contrast against the nav options. Can we override that with something like this in `index.scss`:
```
:root {
	--bs-primary-rgb: 20, 23, 30;
}
```
* The border for inputs is very light and difficult to see against the `$body-bg: #e9ecef;` override in `index.scss`. I think removing that override is fine for now to correct visiblity. We can later do some color theme overrides in `index.scss` using this pattern after the BS imports:
```
[data-bs-theme="light"] {
	--bs-body-bg: green;
	--bs-border-color: red;
}
```

FeralAI (approved):
👍😁

PR #300, opened 2023-06-04
Move invert Y to process() to avoid hotkey issues

FeralAI:
Now that hotkeys are configurable it's possible to cause the hotkeys bound to `Up` or `Down` to flip if invert Y axis is enabled. This is a small change to move invert Y axis out of `Gamepad::read()` and into `Gamepad::process()` to avoid this issue.


arntsonl (approved):
Nice easy fix

PR #317, opened 2023-06-08
Fix button labels and hotkey when Touchpad Share are swapped

teikai1216:
*Fix HotkeyF1 when touchpad and share button are swapped in ps4 mode

*Fix WebConfig button labels when touchpad and share button are swapped in ps4 mode


arntsonl (approved):
Looks like a clean fix for our swapping setting, thank you @teikai1216 !

PR #320, opened 2023-06-09
Setting to treat the dpad as a 4-way joystick

bsstephan:
This masks off the oldest cardinal direction of a diagonal, allowing for fast switching to new cardinal directions on a lever, meant for arcade games which don't benefit from diagonals and would likely benefit from unambiguous, immediate direction change.


bsstephan:
I'm going to add a hotkey action to toggle DDI 4-Way mode as well, but I think I'll wait for #325 to merge first, because my three adjacent branches are all getting to be a bit much to juggle.

arntsonl (approved):
Works great!

PR #322, opened 2023-06-10
Exclude fields from JSON config export

mthiesen:
I extended NanoPb to allow Protobuf fields to be flagged with `(nanopb).disallow_export`. Fields with this flag will now be excluded when converting the config to JSON.

I have also reenabled the `/api/getConfig` endpoint.


arntsonl (approved):
This seems cleaner than removing /getConfig

PR #325, opened 2023-06-12
only trigger a hotkey action once per press

bsstephan:
working on 4-way joystick mode toggling in another branch, it was random whether or not I "landed" on set or unset. I identified this as the hotkey code not properly tracking lastAction.

before this change, while a hotkey was held, the action would be applied every time hotkey() was called, e.g. every millisecond, because lastAction was always NONE. this was only checked on the Y-axis toggle, but is now checked on all hotkey actions, as it also allows us from calling save() unnecessarily.

following this change I no longer experience the ambiguity on where I "land" for hotkeys that toggle a setting

### DIFF
--- a/headers/gamepad.h
+++ b/headers/gamepad.h
@@ -4,6 +4,7 @@
 #include "BoardConfig.h"
 #include <string.h>
 
+#include "enums.pb.h"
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Looks good!

PR #333, opened 2023-06-13
Update to config

TheTrainGoes:


arntsonl (approved):
Nice, automatic bentobox boot image :)

PR #335, opened 2023-06-14
Revert back to global js includes

FeralAI:
This should fix the issue of some browsers not loading the doc site properly. My assumption is the non-working browsers either didn't support ES browser modules or had a timing issue processing the IIFE for the Docsify init.


arntsonl (approved):
Too bad those browsers don't support ES :(

PR #336, opened 2023-06-14
move PS4 TP/Share switch check to getPS4Report

### LINKED ISSUE #329: Move options.switchTpShareForDs4 check to getPS4Report()
As per https://github.com/OpenStickCommunity/GP2040-CE/commit/ddf59c3c345ef0815921e26c1d7679b5a7aff727#commitcomment-117842350, the PS4 TP/Share toggle manipulates the actual pin mapping of S1 and A2, which is going to be an issue for my hotkey PR which uses the button masks directly (and thus encapsulating the feature would mean duplicating the logic there), and I think it's misleading to other parts of the code.

After a quick Discord conversation I think we can move this check to the PS4-specific report, and then S1 always is S1.

I'll try to get to this soon, if no one beats me to it.

bsstephan:
this allows the rest of the gamepad code --- hotkeys, BOOTSEL boot hotkey --- to always treat S1 as S1 and A2 as A2, rather than having to also encapsulate the switch check in order to have consistent behavior. the switch is only for PS4 mode anyway, so we can just stuff it in there.

fixes #329

### DIFF
--- a/src/gamepad.cpp
+++ b/src/gamepad.cpp
@@ -118,14 +118,6 @@ void Gamepad::setup()
 	mapButtonA1  = new GamepadButtonMapping(convertPin(pinMappings.pinButtonA1),	GAMEPAD_MASK_A1);
 	mapButtonA2  = new GamepadButtonMapping(convertPin(pinMappings.pinButtonA2),	GAMEPAD_MASK_A2);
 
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Easy peasy

PR #340, opened 2023-06-16
Dual dir defaults

Pelsin:
This solves the issue with not being to change the dual dir inputs with up/down buttons or setting it to "0"


arntsonl (approved):
Nice, good fix for Dual Directional

PR #343, opened 2023-06-17
Implement localisation

### LINKED ISSUE #285: The web configurator is English only

Yes

Non-English speakers are not able to use the web config.

Under settings, there should be a dropdown to choose supported languages. We should also auto-select the users' language with `i18next-browser-languagedetector`.

Pretend English is number 1 and no one else exists. 🇬🇧 🇬🇧

Yes

Yes. I've started here: https://github.com/SavageCore/GP2040-CE/tree/feature/i18n

I can extract all of the English strings and prepare the app for translators but then we'll need community members to do the rest! Or DeepL 🤷

SavageCore:
Closes #285

Putting this up here now for feedback, busy weekend ahead with Father's Day.  I think I've caught the vast majority of strings but some may have slipped through the cracks.  Dutch translations are only for testing, used DeepL mostly, ik kan niet goed Nederlands spreken... 

Feedback on translation keys very welcome, this is my first time with i18n.

British English just for a laff/testing.


www/src/Pages/HomePage.jsx
```diff
@@ -50,29 +53,29 @@ export default function HomePage() {
...
-			<Section title="System Stats">
+			<h1>{t('pages.home.header.label')}</h1>
+			<p>{t('pages.home.subheader.label')}</p>
+			<Section title={t('pages.home.system-stats.version.header.label')}>
 				<div>
-					<div><strong>Version</strong></div>
-					<div>Current: { currentVersion }</div>
-					<div>Latest: { latestVersion }</div>
+					<div><strong>{t('pages.home.system-stats.version.label')}</strong></div>
+					<div>{t('pages.home.system-stats.version.current.label')}: {currentVersion}</div>
```
Pelsin, line 61: Interpolation might be good to here

Pelsin:
Nice!

I might have some opinions on how to add i18n, not sure they are optimal but have helped me in larger projects.

i18next-http-backend might be overkill here as we only are using it for fetching the default locales location with namespace, i think defining the [resources ](https://www.i18next.com/how-to/add-or-load-translations#add-on-init) config with custom namespaces and separated files makes for an easier structure.

E.g of namespaces:
- Common
- HomePage
- BackupPage
- Components

We should then be able use the hook with: `useTranslation('BackupPage')` 
Another thing that I find easier to work with is a much flatter json for translations:
```json
"pin-mapping-header": "Pin"
```
Instead of:
```json
"pin-mapping": {
		"table": {
			"headers": {
				"pin": {
					"label": "Pin"
				}
			}
		}
	},
```

Might also be nice to add [eslint-plugin-i18next](https://github.com/edvardchen/eslint-plugin-i18next) for non-localized text

www/package.json
Pelsin, line 1: Interesting why the whole file changed, maybe gitattribute

Pelsin, line 1: Interesting why the whole file changed, maybe the gitattributes config didnt work or it's something else

www/src/index.jsx
Pelsin, line 1: Same here, the entire file changed 🤔

www/package.json
SavageCore: Yes, odd. GitKraken correctly displays a partial change in the git history but once it's pushed to GH...

SavageCore:
Thanks for the comprehensive review, I will start going through the changes. eslint plugin is great!

Pelsin:
Looks awesome!
Wondering what the other lord of the code people think about the process of adding new languages/maintaining them.

Regardless, we should be able to merge this with just the base language so the functionality is there when we want to add a new language?

SavageCore:
Finished the rebase with main and the change to namespaces. Probably good to review it again. I was going to test JSX vs JSON lang files regarding bundle size soon.

SavageCore:
Update, [JSX](https://htmlpreview.github.io/?https://gist.githubusercontent.com/SavageCore/f14bfbfbb8b6825f37c7ee0d996c635e/raw/3a95fd2ad8a28d7582dcae776d3766211e74e568/jsx.html) v [JSON](https://htmlpreview.github.io/?https://gist.githubusercontent.com/SavageCore/f14bfbfbb8b6825f37c7ee0d996c635e/raw/3a95fd2ad8a28d7582dcae776d3766211e74e568/json.html). 

Any opinions? JSX is 7% smaller if my calculations are correct.

PR #344, opened 2023-06-17
provide tool to visualize board config from flash dump

bsstephan:
This python script (with instructions in the header) is a simple tool to parse and visualize the protobuf config from a board dump via picotool save.

This may be of use to anyone making significant protobuf changes, or just curious about the storage, I'm submitting a PR just in case anyone else finds it interesting or relevant, or at least worth keeping around.


bsstephan:
As discussed I am starting a new repo for tools. I will add this to https://github.com/OpenStickCommunity/binary-tools once I get a chance.

PR #349, opened 2023-06-18
Reduce binary size

mthiesen:
This PR reduces the size of the firmware binary. In my local tests the new size is 1.373.696 bytes, whereas the previous size was 1.617.408 bytes.

I achieved this through various means:

*   Using the `-fdata-sections`, `-ffunction-sections` and `--gc-sections` compiler flags to remove unused code and data
*   More aggressive compression of the webconfig assets
*   Minimal implementation of the GCC terminate handler, the default implementation is very large because it does a lot (i.e. demangle function name) that we do not need anyway
*   Reduced the size of the `ConfigUtil::fromJSON()` function


arntsonl (approved):
Super easy binary size reduction, thanks!!

PR #350, opened 2023-06-18
Reduce memory usage

mthiesen:
This PR reduces the memory usage (in webconfig mode) from 138 kb to 109 kb.

I achieved these improvement with the following changes:

*   Use `inline const` or `constexp` instead of `static const` in headers. The latter creates a new instance of the constant for each translation unit that includes the header, whereas the former creates a single instance for all translation units.
*   Converted global variables to local variables when possible
*   Converted constants with allocating types (i.e. `std::vector`, `std::string`) to non-allocating types (i.e. `C arrays`, `C strings`). The latter can be used directly from flash, whereas the former need to be copied to an allocated memory block first.

### DIFF
--- a/headers/helper.h
+++ b/headers/helper.h
@@ -13,25 +13,6 @@
 // GP2040-CE Version (32 character limit)
 #define GP2040VERSION "v0.7.2"
 
-const std::string BUTTON_LABEL_UP = "Up";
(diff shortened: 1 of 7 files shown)


arntsonl (approved):
Nice, those are simple ways to reduce memory!

PR #354, opened 2023-06-19
Add multimedia key options to Keyboard mode

deeebug:
This update introduces multimedia key inputs to the Keyboard mode. The keys can be mapped from the usual keyboard mapping page. And the following keys are available:
- Next Track
- Prev Track
- Stop
- Play/Pause
- Mute
- Volume Up
- Volume Down

For this, the Keyboard USB descriptor has been updated and other required changes have also been made to send input reports to proper report IDs.


arntsonl (approved):
Tested and working great!

PR #358, opened 2023-06-21
Initial support for LKP (my capacitive slider controller module)

dogtopus:
[LKP](https://github.com/Project-Alpaca/LKP) is a capacitive slider controller module inspired by certain elements in some arcade rhythm game controllers such as [that Miku game](https://commons.wikimedia.org/wiki/File:Hatsune_Miku_Project_DIVA_Arcade_Future_Tone_Version_A_20150327.jpg). It uses I2C interface with an interrupt pin to communicate with the host and provides digital ON/OFF reading of 32 individual capacitive touch sensors. This PR adds initial support for this module in GP2040.

Currently it is hardcoded to emulate the Hori analog stick protocol used in the official Project DIVA Future Tone/Mega Mix controller (documented [here](https://www.youtube.com/watch?v=dwpU5-l3zz8)) but I'm looking into adding other modes in the future, as well as hotkeys that switch between several modes at run-time. Possible interesting modes including emulating only analog stick left/right pushes (similar to what [Divaller](https://www.dj-dao.com/en/divaller) did) or binding touch sensors to key presses.

(~~Currently this is in draft because there is a regression on main branch that prevented addon settings from saving and I'm waiting for it to be fixed so I can test the rebased version.~~ ~~Fixed locally. Will test the changes here later. On my local development branch based on older code the PR seems to work fine.~~ Fully tested and ready.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,6 +139,7 @@ src/addons/keyboard_host.cpp
 src/addons/i2canalog1219.cpp
 src/addons/jslider.cpp
 src/addons/i2cdisplay.cpp
+src/addons/i2clkp.cpp
 src/addons/neopicoleds.cpp
 src/addons/playernum.cpp
 src/addons/playerleds.cpp
--- a/headers/addons/i2clkp.h
+++ b/headers/addons/i2clkp.h
@@ -1,0 +1,102 @@
+#ifndef _I2CLKP_H
+#define _I2CLKP_H
+
+#include <hardware/i2c.h>
+
+#include "gpaddon.h"
+#include "GamepadEnums.h"
+
+#define I2CLKPName "I2CLKP"
+
+#ifndef LKP_ENABLED
+#define LKP_ENABLED 0
+#endif
+
+#ifndef LKP_I2C_BLOCK
+#define LKP_I2C_BLOCK i2c0
+#endif
+
+#ifndef LKP_INTERRUPT_PIN
+#define LKP_INTERRUPT_PIN -1
+#endif
+
+#ifndef LKP_ADDR
+#define LKP_ADDR 0x08
+#endif
+
+#ifndef LKP_SDA_PIN
+#define LKP_SDA_PIN -1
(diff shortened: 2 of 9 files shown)


arntsonl:
Thank for submitting this @dogtopus !  What kind of hardware do we need to test this one? We can order off Amazon or AliE or anywhere.

dogtopus:
@arntsonl Unfortunately it's not for sale at this moment. It is completely open source so you could theoretically build one yourself but it takes time and work.

Though I could be the maintainer for this part of the code if you wish.

arntsonl:
Hi @dogtopus , so our policy right now for community hardware is that one of us needs to be able to test it. So for example, the SNES2USB add-on. I was sent a unit to test and was able to verify the pins are correct, as well as catalog a bug with start-up in our add-ons. The Wii Extension I was able to grab a wii to i2c adapter off Amazon, TheTrain has several of the closed boards we support. We are absolutely not opposed to adding more functionality and add-ons, but we need to make sure more than one person can support and test a feature we add to main. This also ensures our development will not drop off on a particular feature if one developer needs a break or goes on a hiatus, and we don't rely on a single user for our support channel.

So, here's what I'm going to suggest as this looks like a super cool feature and I would definitely think folks would use it in our firmware:

- Create a fork (or branch directly) of GP2040-CE and keep it up to date with our latest main
- Once your LKP is ready for beta-testing and you have some units on hand, let us know and we'll happily buy one from you for dev and testing
- Join our Discord and let us know if there's anything else you want to see in the project!!

dogtopus:
@arntsonl LKP uses the old RepRap distribution model, which is a sugar coated way of saying "all manufacturing data available and if anyone is interested in it, they can create a group and discuss about making a small batch for themselves". Unfortunately this makes it quite hard, if not impossible, to just "buy one from somewhere".

I need to figure something out for this to work. I guess worst case I could always just mail one of the pre-assembled test units to one of the GP2040 devs instead.

PR #359, opened 2023-06-21
Fix regression on addon settings saving

dogtopus:
This was introduced in 031ba64.

Settings were not correctly validated since the pin validator is used to validate boolean values, which caused it to always fail silently.


arntsonl (approved):
Good catch, thank you!

PR #366, opened 2023-06-22
BUGFIX: don't check lastAction for non-toggle, non-save hotkey actions

### LINKED ISSUE #360: HOTKEY_HOME_BUTTON doesn't do anything because I broke it
HOTKEY_HOME_BUTTON no longer does anything perceivable in `main`, I'm pretty sure because of 9c7ed0281fd8950db81ca5bbe0b97d4440fe5a70 --- I think in reality it is actually being pressed for 1ms and then disappears.

Probably what needs to happen is hotkey actions that modify `state.buttons` shouldn't be subject to the `action != lastAction` check. This will take some rejiggering in the code.

Note that this issue also blocks the fix in #342 as it also intends to set `state.buttons`.

bsstephan:
the top-level check was actually somewhat redundant now that I fixed lastAction, since the toggles checked it anyway, but we can still use a similar approach to not save unless we need to.

but now, we also don't interfere with e.g. HOTKEY_HOME_BUTTON, which *should* always take effect as long as the hotkey is pressed, even if we've done it in past cycles --- but it shouldn't save, so don't request it.

fixes #360

### DIFF
--- a/src/gamepad.cpp
+++ b/src/gamepad.cpp
@@ -300,40 +300,48 @@ void Gamepad::hotkey()
  */
 void Gamepad::processHotkeyIfNewAction(GamepadHotkey action)
 {
-	if (action != lastAction) {
-		switch (action) {
-			case HOTKEY_NONE              : return;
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Looks good!

PR #368, opened 2023-06-24
Add missing translations for KeyboardMapper Component

SavageCore:
Does what it says on the tin as well as correcting the tab spacing


arntsonl (approved):
Makes sense, looks good!

PR #372, opened 2023-06-24
Fix compile error with some GCC versions

mthiesen:
This PR fixes the compile error discovered by @chiyu521. This only happens with some versions of GCC, I am not sure which ones. However, the fix should work with all reasonably recent versions of GCC.


arntsonl (approved):
Interesting! I think I saw this on one of my machines.

PR #373, opened 2023-06-24
resolve #361

### LINKED ISSUE #361: HOTKEY_INVERT_X_AXIS doesn't do anything, never did?
Looking at the git history, it seems `HOTKEY_INVERT_X_AXIS` has always just done a `break` and accomplished nothing. I think it could work just like the the other axis, so it was probably just an oversight/thinko.

NickGuyver:
Resolve issue #361


arntsonl (approved):
Nice, seems like a good little fix

PR #374, opened 2023-06-24
PS4 key management overhaul

### LINKED ISSUE #363: Rework the PS4 key importing logic

Yes.

I'm always frustrated when PS4 key import is so convoluted and fails at random places without a clear error message indicating where and how should I fix it.

Rework the importation to:

- [x] Use mbedTLS for importation and verification from a single raw DS4 key binary file (1424 bytes, exactly like how it's laid out on an official JDM-001 controller after deobfuscation).
  - PassingLink was unable to do this because it aims at supporting extremely low RAM devices like STM32 blue pill. This is not a concern for us.
- [x] Front-end only takes a single file, uploads it to GP2040. Then GP2040 takes over, do point 1, save the key if everything is good and return the result.
- [ ] (Optional, maybe as a second step) Display at least the serial number of the key in flash on the PS4 addon settings page.

(Quick reference of DS4 key format mentioned above (named `DS4FullKeyBlock`), as implemented in my [DS4 USB function project](https://github.com/dogtopus/ffsds4/blob/bff7c98ceafab78f801f060b8867e6a950c25ffb/ffsds4lib/ds4.py#L478-L529))

Fix the current import, but again there's too many moving parts on the front-end with a lot of strange back and forth conversions.

Yes

Yes

dogtopus:
This streamlines the PS4 authentication key management and usage.

- Split PS4 authentication and key validation to its own library.
- Use all standard mbedTLS APIs for authentication, as well as key validation.
- Support both a single DS4 key block file (same format as the one dumped from JDM-001 motherboard) and the old 3-file PassingLink-style key format.
- Key validation is now done (in a pretty clean way) on board instead of on front-end and completes within reasonable time (~4-5 seconds). This gives extra assurance that the key will be more likely to work on-board and cut down the overall system complexity.
- Front-end now shows more diagnostic information. No more "I clicked the button it doesn't do anything" complains.

Closes #363.


dogtopus:
I talked to @TheTrainGoes for a bit and they think that we shouldn't drop the 3 files import. I'll look into that. Meanwhile this will stay drafted for now.

arntsonl:
Yeah I wrote this code, I'm lucipher in discord btw.

In terms of importing the 3 files, I did it this way for a few reasons. The biggest one is that we can use the computer/web to handle pulling the modulus out of the private key rather than mbedtls. I was having HUGE timeouts on the rp2040 when I tried to do this on the board, to the point where my upload POST would just fail. But I also didn't have pico-mbedtls until after we started testing this, which brought the RSA signing times way down.

In terms of code so far, a couple of things.

1. I'd like to keep all of our licensing to the project rather than individuals. Right now they are set to FeralAI's name, but I'd like to go even further out and change our licensing to OpenStickCommunity as we are always open and free.

2. Let's keep code like this:  tl;dr: TBH IDRC just use it LMAO.   out of main just for cleanliness sakes. I've let a few things sneak in of my own, but I try to pull them when I see them now that we have 10+ developers on here actively committing changes.

dogtopus:
Now the 3-file key import is added back, this should be ready for a second review.

arntsonl:
Alright, I got some feedback from everybody and looked over the code myself. I'm going to push the PR back to you because we have some bigger design feedback requests to move this over to main:

1. Let's not store the ds4key in the protobuf storage, this adds an extra 1.4kb bytes for data we already have through the RSA prime numbers.
2. We also want to move the jedi key blob handling to the web config as it doesn't make sense to rely on mbedtls to handle this for us when we can do it on the web side
3. We can keep ds4crypto but you'll have to modify it to generate the RSA keys using the modulus and primes stored in protobuf.

Let me know if you have any questions, you got me on Discord.

dogtopus:
Closing this now since we can't reach an agreement on the technical details.

Though feel free to use it as a reference for things going forward.

PR #375, opened 2023-06-25
PS4 Key Web-UI Fix

arntsonl:
Fix for the Web UI to accept PS4 keys again ('jsbn' broke in latest jsx changes)


arntsonl:
Should have done this in the first place, replaced jsbn with BigInt and used a third-party library for mod inversion.

PR #377, opened 2023-06-29
Refactor PS4 auth

dogtopus:
This is a smaller scale refactor than #374.

Notable changes:
- Frontend now uses Promise.all to load all required files.
- RSA key loading from config and identity exporting after signing are now done in mbedTLS API.
- Cut down redundant RSA factors in config storage. Now it only saves N, E, P and Q to the config and calculates the rest of the factors during setup(). This should save around 900 bytes of config storage.

### DIFF
--- a/headers/addons/ps4mode.h
+++ b/headers/addons/ps4mode.h
@@ -15,23 +15,14 @@
 
 class PS4ModeAddon : public GPAddon {
 public:
-    virtual bool available();
+	virtual bool available();
 	virtual void setup();       // TURBO Button Setup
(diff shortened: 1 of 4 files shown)


arntsonl (changes requested):
One change on the if check, but it looks good. Changing the file read to a Promise await makes sense, I'll have someone with a PS4 make sure the logic all works.

arntsonl (approved):

arntsonl (approved):
This refactor looks much better, saves us a good chunk of bytes in running memory and protobuf

PR #379, opened 2023-06-29
Show a loading spinner when performing WebApi `GET`s

SavageCore:
Adds a loading spinner to `Section`s when a `GET` is being performed and the data isn't ready yet.


arntsonl (approved):
This is a really great useability feature, I really like it!

Issue #14, opened 2022-12-21
Wrong LED order when inverting DPAD mode


TheTrainGoes:
when using inverted DPAD if the down button is pressed the LED for the up button is illuminated and visa versa
video of bug:
https://photos.app.goo.gl/UAk4NvH8KEwhtYp29

To elaborate on this my suspicion is that when the invert DPAD shortcut is used it is not also switching the assigned RGB LEDs order.

```A toggle is available to invert the Y-axis input of the D-pad, allowing some additional input flexibility. To toggle, press START + HOME + RIGHT. This is a temporary hotkey mapping for this feature, so keep an eye on updated releases for this to change.```

TheTrainGoes:
This has been completed.

PR #381, opened 2023-06-30
resolve issue 14

NickGuyver:
fix for "Wrong LED order when inverting DPAD mode" https://github.com/OpenStickCommunity/GP2040-CE/issues/14


TheTrainGoes:
I have tested this and it is working - thanks @NickGuyver

arntsonl (approved):
Very tidy fix, thanks!

PR #384, opened 2023-07-02
Ability to switch between action/direction button profiles

bsstephan:
Working:
* config section defines up to three three alternative action button slots
* hotkey actions to switch to them
* the actual application of profiles (2-4 overlay the total core mappings in the base settings, which is profile 1)
* config option to store selected profile
* exposing the alt mappings in the API
* web configurator profile setting page w/i18n
* actual board saves and intended behavior confirmed and checked with binary-tools
* above for directionals

### DIFF
--- a/www/src/Services/WebApi.js
+++ b/www/src/Services/WebApi.js
@@ -25,6 +25,49 @@ export const baseButtonMappings = {
 	Fn:    { pin: -1, key: 0, error: null },
 };
 
+export const baseProfileOptions = {
+	alternativePinMappings: [{
+		Up:    { pin: -1, key: 0, error: null },
+		Down:  { pin: -1, key: 0, error: null },
+		Left:  { pin: -1, key: 0, error: null },
+		Right: { pin: -1, key: 0, error: null },
+		B1:    { pin: -1, key: 0, error: null },
+		B2:    { pin: -1, key: 0, error: null },
+		B3:    { pin: -1, key: 0, error: null },
+		B4:    { pin: -1, key: 0, error: null },
+		L1:    { pin: -1, key: 0, error: null },
+		R1:    { pin: -1, key: 0, error: null },
+		L2:    { pin: -1, key: 0, error: null },
+		R2:    { pin: -1, key: 0, error: null },
+	},{
+		Up:    { pin: -1, key: 0, error: null },
+		Down:  { pin: -1, key: 0, error: null },
+		Left:  { pin: -1, key: 0, error: null },
+		Right: { pin: -1, key: 0, error: null },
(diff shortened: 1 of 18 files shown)


Pelsin:
When refreshing the page another request gets trigger and sets `loading` to `false` when completed, this happens after `getProfileOptions` has started.

When navigation from another route the other request that set loading to false has already happen, now the only request that gets trigger is `getProfileOptions` that never sets `loading` to `false` on completion.

In essence this is because we are using one property `loading` to indicate if any request is flight. Which can get kinda wonky as it will become a race condition.

When the time present itself i'll rewrite the state handling loading.

www/src/Services/WebApi.js
```diff
@@ -235,6 +266,25 @@ async function setPinMappings(mappings) {
...
+	try {
+		const response = await axios.get(`${baseUrl}/api/getProfileOptions`);
+		let profileOptions = { ...baseProfileOptions };
+		response.data['alternativePinMappings'].forEach((altButtons, index) => {
+			for (let prop of Object.keys(altButtons))
+				profileOptions['alternativePinMappings'][index][prop].pin = parseInt(
+					response.data['alternativePinMappings'][index][prop]
+				);
+		});
+		return profileOptions;
```
Pelsin, line 281: To solve the loading spinner issue, set loading to false before the return:
`setLoading(false);`

src/config_legacy.cpp
```diff
@@ -670,6 +670,7 @@ bool ConfigUtils::fromLegacyStorage(Config& config)
         }
         SET_PROPERTY(gamepadOptions, invertXAxis, legacyGamepadOptions.invertXAxis);
         SET_PROPERTY(gamepadOptions, invertYAxis, legacyGamepadOptions.invertYAxis);
+        SET_PROPERTY(gamepadOptions, profileNumber, 1);
```
mthiesen, line 673: You should not change `config_legacy.cpp`. The file only contains code to perform the config conversion from pre-Protobuf to Protobuf. Default initialization is already performed in `ConfigUtils::initUnsetPropertiesWithDefaults()`.

mthiesen (dismissed):

src/config_legacy.cpp
bsstephan: You are correct, I will remove that.

bsstephan: @mthiesen - I've corrected that and pushed a squashed rebase. I think this is ready for final review and merge.

arntsonl (approved):
Looks good!

arntsonl:
Issues were fixed, there is one new issue that we'll make a Issue# for but might be related to react.

PR #387, opened 2023-07-04
Correctly import isNil

Pelsin:
The isNil was incorrectly imported, uploading **might** have given false positives or not worked at all prior to this change.
This also reduces the size of the bundle by about 80kb.


SavageCore (approved):

arntsonl (approved):
Nice easy fix

PR #388, opened 2023-07-04
Remove unused dependencies

Pelsin:


SavageCore (approved):

arntsonl (approved):
Nice, yeah with the PS4 changes we don't need to calculate the mod inverse for the RSA keys. Good catch!

PR #392, opened 2023-07-05
Only show available analog pins on addons page

Pelsin:
- Options for analog pins are now filtered by used pins
- Fixed some linting issues
- Added translation for "none"


arntsonl (approved):
Nice thanks for fixing this!!

PR #393, opened 2023-07-05
Options should not be filtered as chosen pin cannot be shown

SavageCore:


Pelsin:
Nice! Thanks for fixing my mistakes <3

Pelsin (approved):

SavageCore:
> Nice! Thanks for fixing my mistakes <3

Team work makes the dream work and all that ;)

PR #394, opened 2023-07-05
fix HOTKEY_CAPTURE_BUTTON

bsstephan:
The capture button hotkey now presses S1 (unless we're in PS4 mode and the toggle is on, in which case it presses A2).

Note that this hotkey does something even on the input modes that don't actually have a capture functionality, for clarity's sake.


arntsonl (approved):
Looks good!

PR #399, opened 2023-07-07
analog drift auto calibration

NickGuyver:
Automatically compensate for drifting analog thumb sticks based on center drift at start

### DIFF
--- a/headers/addons/analog.h
+++ b/headers/addons/analog.h
@@ -64,6 +64,7 @@ public:
 	virtual void preprocess() {}
     virtual std::string name() { return AnalogName; }
 private:
+	uint16_t map(uint16_t x, uint16_t in_min, uint16_t in_max, uint16_t out_min, uint16_t out_max);
 };
 
 #endif  // _Analog_H_
--- a/src/addons/analog.cpp
+++ b/src/addons/analog.cpp
@@ -12,6 +12,10 @@
 #define ADC_MAX ((1 << 12) - 1)
 #define ANALOG_CENTER 0.5f // 0.5f is center
 #define ANALOG_MAX 1.0f    // 1.0f is max
+uint16_t adc_1_x_center = 0;
+uint16_t adc_1_y_center = 0;
+uint16_t adc_2_x_center = 0;
+uint16_t adc_2_y_center = 0;
 
 bool AnalogInput::available() {
     return Storage::getInstance().getAddonOptions().analogOptions.enabled;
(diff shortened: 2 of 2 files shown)


headers/addons/analog.h
```diff
@@ -64,6 +68,7 @@ class AnalogInput : public GPAddon {
 	virtual void preprocess() {}
     virtual std::string name() { return AnalogName; }
 private:
+	uint16_t map(uint16_t x, uint16_t in_min, uint16_t in_max, uint16_t out_min, uint16_t out_max);
```
mthiesen, line 71: This can be made `static`.

src/addons/analog.cpp
```diff
@@ -13,6 +13,11 @@
 #define ANALOG_CENTER 0.5f // 0.5f is center
 #define ANALOG_MAX 1.0f    // 1.0f is max
 
+uint16_t adc_1_x_center = 0;
```
mthiesen, line 16: Is there a particular reason why these variables are placed in the global scope? If at all possible these should be member variables of `AnalogInput`.

src/addons/analog.cpp
```diff
@@ -33,10 +38,31 @@ void AnalogInput::setup() {
...
+		if ( isValidPin(analogOptions.analogAdc2PinY) ) {
+			adc_select_input(analogOptions.analogAdc2PinY-26); // ANALOG2-Y
+			adc_2_y_center = adc_read();
+		}
+	}
 }
 
 void AnalogInput::process()
 {
+
```
mthiesen, line 65: This empty line was probably added by accident.

src/addons/analog.cpp
```diff
@@ -46,8 +72,30 @@ void AnalogInput::process()
     float deadzone = analogOptions.analog_deadzone / 200.0f;
 
     if ( isValidPin(analogOptions.analogAdc1PinX) ) {
-        adc_select_input(analogOptions.analogAdc1PinX-26); // ANALOG-X
-        adc_1_x = ((float)adc_read())/ADC_MAX;
+        adc_select_input(analogOptions.analogAdc1PinX-26); // ANALOG1-X
```
mthiesen, line 75: There is a lot of code duplication in this function, you basically do the same for all axes. You should extract the common code into a new function and call this four each axis instead.

src/config_legacy.cpp
```diff
@@ -245,6 +245,7 @@ namespace ConfigLegacy
         uint8_t analogAdcPinY;
         uint8_t forced_circularity;
         uint8_t analog_deadzone;
+        uint8_t auto_calibrate;
```
mthiesen, line 248: You should not change `config_legacy.cpp` at all. These `struct`s represent the final state of our config storage before the migration to Protobuf. Changing their structure will break the migration from pre-Protobuf config to Protobuf config.

src/config_utils.cpp
```diff
@@ -326,6 +326,7 @@ void ConfigUtils::initUnsetPropertiesWithDefaults(Config& config)
     INIT_UNSET_PROPERTY(config.addonOptions.analogOptions, analogAdc2Invert, ANALOG_ADC_2_INVERT);
     INIT_UNSET_PROPERTY(config.addonOptions.analogOptions, enabled, !!FORCED_CIRCULARITY_ENABLED);
     INIT_UNSET_PROPERTY(config.addonOptions.analogOptions, analog_deadzone, DEFAULT_ANALOG_DEADZONE);
+    INIT_UNSET_PROPERTY(config.addonOptions.analogOptions, enabled, !!AUTO_CALIBRATE_ENABLED);
```
mthiesen, line 329: You overwrite the value of `analogOptions.enabled` that was just set two lines above. Typo?

mthiesen (changes requested):

NickGuyver:
Thanks for the review @mthiesen , I went ahead and removed some other things I had added to the legacy from previous additions.

mthiesen:
Looks good to me now! Thanks @NickGuyver!
Still, somebody with an actual analog stick device should try this out before we merge.

mthiesen (approved):

PR #401, opened 2023-07-08
Add Flatbox Rev5 RGB

SavageCore:
Test build: https://github.com/OpenStickCommunity/GP2040-CE/suites/14156305162/artifacts/793156062


SavageCore:
I assume this works fine for v2 and the user will just need to set LEDs per button to 2?

TheTrainGoes:
That's correct.  Rev1 would also need the PLED numbers changed on it.  

V1 is depreciated now so I'm fine just hosting v2.

SavageCore:
@TheTrainGoes OK makes sense. Updated to 2 LEDs per button, PLED pins are 12-15, is that right?

TheTrainGoes:
PLED pins are 24, 25, 26 and 27

TheTrainGoes:
Correction - LED # not PIN number.

TheTrainGoes (approved):

PR #403, opened 2023-07-08
add a BOOTSEL reboot hotkey to the "secret" hotkeys

bsstephan:
Simple enough. I wrote this because I usually want to bounce between gamepad and boot USB mode, and less so webconfig mode. I put it on S1 + B3 + B4 to match the behavior of the webconfig reboot.

Aside from adding the mask check, most of this is just renaming the struct to reflect that it has two hotkeys in it now.

This is ready to go, AFAIK, but I'm marking it as a draft because I don't think there's any reason to rush it for 0.7.3.

### DIFF
--- a/headers/gp2040.h
+++ b/headers/gp2040.h
@@ -23,17 +23,18 @@ private:
     Gamepad snapshot;
     AddonManager addons;
 
-    struct WebConfigHotkey {
-        WebConfigHotkey();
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Looks good! It'll be nice to have a few secret shortcuts :D

PR #321, opened 2023-06-10
Add forced circularity option and minor fix to analog.cpp

NickGuyver:
Add an analog configuration option to force perfect circularity on analog joystick. This video does a good job of showing why it should be an option - (https://www.youtube.com/watch?v=1P1mItHTKJw)

Nest the deadzone if statements inside the statements where their respective axis is validated.

PR #404, opened 2023-07-09
Removed harmful changes to config_legacy.cpp

mthiesen:
PR #321 changed one of the `struct`s in `config_legacy.cpp`. We can never do that, our pre-Protobuf migration process relies on these `struct`s not changing.

I have undone the changes and added a comment at the top of the file to prevent such errors in the future.

This PR **must** be merged before the next release. Otherwise, users may experience failed configuration migrations.


TheTrainGoes:
Merged - thanks @mthiesen

PR #417, opened 2023-07-20
Add-Ons UI update

matan-makes:
image

In this PR, I've done the following:

- Added a utility Sass file with a few breakpoint mixins to use for common screen sizes. Not accounting for mobile because I can't think of anyone using their web configurator on a phone, though I guess it is possible.
- Broke up the add-ons page into CSS grid-based columns, can change the number of columns per breakpoint for future updates. I felt it was too much on a larger screen, and I tend to keep my browser at 1080p.
- Updated the PS4 Mode module to be a little more... hip.

In my opinion, this page could use a full refresh. Side menu that has all add-on options that, when clicked, load only that add-on configuration to be changed. Add-ons broken up into individual components rather than having a massive file.

### DIFF
--- a/www/src/Pages/AddonsConfigPage.jsx
+++ b/www/src/Pages/AddonsConfigPage.jsx
@@ -1,5 +1,5 @@
 import React, { useContext, useEffect, useState } from "react";
-import { Button, Form, Row, FormCheck } from "react-bootstrap";
+import { Button, Form, Row, FormCheck, Alert } from "react-bootstrap";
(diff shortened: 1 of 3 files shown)


matan-makes:
of course after all of this, I see that my formatter.... formatted....

arntsonl:
Looks super clean!!! We'll get this merged in after our 0.7.3 release which should be early next week.

matan-makes:
Closing in favor of re-doing with no autoformatting on :)

PR #420, opened 2023-07-21
JSlider single-pin fix

arntsonl:
Modifying jslider to fix single-pin assignment issue.


arntsonl:
Confirmed working by @TheTrainGoes

PR #430, opened 2023-07-27
adding hotkey options for l3, r3 and touchpad buttons

ccristob:


henrebotha:
Why not add Start? Then we have all 6 auxiliary buttons available.

bsstephan:
Other than the above notes, this looks as expected to me. I'm not sure if anyone would ever need a Start hotkey, as they couldn't even get into the web config without a real Start. I personally think that if the need arises, it'll be trivial to add then.

arntsonl (dismissed):
Looks clean! These are very useful for some upcoming things I want to make :)

arntsonl (approved):

PR #436, opened 2023-07-30
Format all web code

Pelsin:
No changes in code/functionality except the .editorconfig.
This is purely for developer experience, also want to make it clear this is just a suggestion on format according to prettier+editorconfig. Just wanted to normalize so we use the same formatting in all files.

This will help issues like what happened here: https://github.com/OpenStickCommunity/GP2040-CE/pull/417
Also thought it was a good idea to do it now as there are only a few PRs up touching the web part.

Ran prettier to format the files in the www folder.
`npx prettier --write "{www/src,www/server}/**/*.{tsx,ts,jsx,js,json,scss}"`

Changed .editorconfig:
- handle EOL as lf, so it matches .gitattributes
- js/ts files use single quotes

Added "format" npm script for those not running a prettier plugin.


SavageCore (changes requested):
I'd prefer `singleQuote`: `true` but otherwise, very nice and very welcome!

Should we add [husky](https://github.com/typicode/husky) to enforce linting on commit? 

`lint` script should be added to the `package.json` and we should then extend `node.js.yml` action to run lint tests.

Pelsin:
> I'd prefer `singleQuote`: `true` but otherwise, very nice and very welcome!
True, i also prefer single quotes. I've changing it in editorconfig for js/ts files. Will check with the other if someone objects :)

> Should we add [husky](https://github.com/typicode/husky) to enforce linting on commit?
> `lint` script should be added to the `package.json` and we should then extend `node.js.yml` action to run lint tests.

Sounds like a great idea, had a short discussion with feralAi about the linting when we moved from CRA to vite.
There currently is a lint script in package.json, there are some errors that we wanted to clean up before making linting part of the pipeline.
Husky and adding to pipeline feels like a really nice next step, in another branch thought!

Thanks for checking the PR out @SavageCore!

SavageCore (approved):

arntsonl (approved):
Looks awesome!

Issue #395, opened 2023-07-06
Key->Input config results in incorrect Key->Input controller mapping


InfraredAces:
### Prerequisites

Please check the following before posting an issue / bug report.

- [ ] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

* Firmware Version: [CMake #1066](https://github.com/OpenStickCommunity/GP2040-CE/actions/runs/5456826754)
* Name of device: RP2040 Feather USB A Host
* Link to where you purchased the device: https://www.adafruit.com/product/5302
* Is this a custom build?: No
* Operating System: 6.4.1-arch1-1
* Browser: Chromium

### Expected Behavior

Use the Keyboard Host Configuration Addon with the RP2040 Feather USB A Host. Plug a keyboard into the USB A host port, and use the configured mappings from the webconfig, resulting in expected inputs.

Intended Configuration

![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/e789a85d-5a6a-4da8-a23c-f7b396abcaad)

### Current Behavior

Use the Keyboard Host Configuration Addon with the RP2040 Feather USB A Host. Plug a keyboard into the USB A host port, and use the configured mappings from the webconfig, resulting in unexpected inputs. Requires swapping inputs as shown below to result in intended keyboard layout

Necessary Configuration for intended layout

![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/a7b2d28e-843b-40b8-b2c1-6ab1b2734a84)

### Steps to Reproduce

1. Flash RP2040 Feather USB A Host with Pico .uf2 file from [CMake #1066](https://github.com/OpenStickCommunity/GP2040-CE/actions/runs/5456826754)
2. Press and hold GPIO 17 (located on USB Host port) and reset to go into WebConfig
3. Disable addons and remap pins appropriately to free up GPIO 16, 17, and 18 for the USB A Host

![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/56cbb33e-f706-46da-bb28-1a4ad4e8b528)

4. Enable Keyboard Host Configuration
[...]

PR #438, opened 2023-07-31
Fix keyboard host mapping

deeebug:
This should address #395.
Thanks to @InfraredAces for thorough testing and a very descriptive writeup about the issue!

The fix is very simple and a mistake from my side, sorry for that.
Edit: The addon has been enabled for testing purpose. It can be disabled if there are difficulties in making the merge.

Please check and merge, thanks!


InfraredAces:
I've flashed my Waveshare board with [this build](https://github.com/OpenStickCommunity/GP2040-CE/actions/runs/5717819255) and set the appropriate data pins so that the Keyboard Host Addon is working with a keyboard plugged into the host port.

The action buttons were fixed so that the labels match the proper input. However, with the function buttons A2 and S1 are reversed and require assignments for the other to work properly.

## Example:
### Desired Inputs and Intended Mapping
- A2 - 2
- S1 - 4

### Necessary mapping to get above
- A2 - 4
- S1 - 2

deeebug:
This is a different issue unrelated to the Keyboard Host addon. The labels get swapped incorrectly. I've pushed fixes for these.

InfraredAces:
This does resolve the issues I was having initially with the PR

arntsonl (approved):
Re-enables keyboard host add-on, looks good!

PR #442, opened 2023-08-04
add a hotkey action to reboot the controller

bsstephan:
Simple way to change modes, primary use case being you powered on your controller before remembering you need to switch modes, and want to switch to the proper mode without getting up or unplugging the cable or whatever --- just press your assigned hotkey while holding the normal input mode button and you'll reboot into that mode (because hey, why bother burning an unplug/plug count on your connector, if you don't have to).


arntsonl (approved):
Perfect! That's a really good FN addition

PR #445, opened 2023-08-06
Update to the basic board configurations

TheTrainGoes:
Updates the Pico, Waveshare RP2040 Zero and Sparkfun Pro Micro RP2040 to disable RGB and OLED display by default.


configs/Pico/BoardConfig.h
```diff
@@ -184,7 +184,7 @@
 // 4 - `NOSPLASH` - This will not display a splash screen on boot
 // Special note - All of the splash screen images can be changed via `include/bitmaps.h`
 
-#define HAS_I2C_DISPLAY 1
+#define HAS_I2C_DISPLAY -1
```
bsstephan: This is used for a bool, so it should be 0, not -1.

bsstephan: There are a couple other BoardConfig.h files that make this mistake, and even `headers/addons/i2cdisplay.h` makes the mistake. I think we should fix `i2cdisplay.h` and remove all board configs that have it defined to 0 (or -1 erroneously) for sake of simplicity.

bsstephan (approved):
Looks good in https://github.com/OpenStickCommunity/GP2040-CE/pull/445/commits/c969f0bd2b65d21522c4eabb9e7529af2da9beff

configs/Pico/BoardConfig.h
TheTrainGoes: Thanks - just saw the comment pop up here :). Fixed!  

We should go through and check the other configs for this

Issue #283, opened 2023-05-25
Add-ons with Input Modifiers fire AFTER the boot selection (fix this)


arntsonl:
### Prerequisites

Please check the following before posting an issue / bug report.

- [x] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: 0.7.1
* Name of device: All Devices
* Link to where you purchased the device: *
* Is this a custom build?: *
* Operating System: *
* Browser: *

### Expected Behavior

The Add-On system is currently not setup to handle input modifiers that fire before the boot selection process. This means that add-ons such as the Wii Extension and SNES/NES pad cannot trigger our web config or other hot keys during boot. We will want to move the logic so that add-ons can load AND run their first pre/post process before the boot actions are validated.

### Current Behavior

Add-ons cannot modify the boot state and therefore cannot handle situations like the SNES 2 USB adapter. In this case, the SNES pad cannot put the GP2040-CE firmware into web config mode.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Enable SNES Mode add-on
2. Using a device such as the SNES 2 USB or KB2040, hold the S2 or any other hotkey button on start-up
3. Observe the GP2040-CE device does not go into web config or any other mode using an add-on device

### Screenshots & Files

None for now

PR #446, opened 2023-08-07
move add-ons before bootsel mode

NickGuyver:
Move the add-ons to load and read before the bootsel mode detection.

Fix issue with SNESpad not reading within start(), which caused held down buttons to always read 0 until repressed.

https://github.com/OpenStickCommunity/GP2040-CE/issues/283


arntsonl:
Just need some conflicts resolved on gp2040.cpp and this is ready to pull in!

arntsonl (approved):
Works great, thank you for updating this!

PR #448, opened 2023-08-07
ADD: Add-on Input Mode Selection Slider

HoboDev:
Allows up to 3 modes to be selected with a slider.


TheTrainGoes:
This is neat!  

I am getting fairly frequent lockups on my unit if I slide too quickly.  Any idea what might be causing that?

HoboDev:
I've experienced that the display locks up but not the unit itself. How is it in your case?
I don't really know what's causing it. But I can recommend to increase the waiting time after disconnect from 600ms to 1000ms.

TheTrainGoes:
@HoboDev just to add to this you are correct - it is just the screen and LEDs that are freezing.  Still an issue but not as bad.

Animations would be good in any case :)

bsstephan:
@TheTrainGoes or @HoboDev - when the freeze occurs, do any of the addons still work? I wonder if somehow core1 is deadlocking.

HoboDev:
One more question. Is the freeze temporary like the result of the 600ms waiting time? Or does the screen lock up until reboot?

TheTrainGoes:
Apologies @HoboDev I have been very bust with recent releases.  

Will circle back to this next week when I can.

arntsonl:
Functionality wise, this is something we already support with a reboot + holding B1-B4. With the addition of a Fn key reboot shortcut, I'm going to close this PR.

For code clean-ups and cleanliness, feel free to add another PR. Also feel free to chat with us on Discord for ideas and feedback!

HoboDev:
Well on Discord someone encouraged the PR after I was talking about a feature like this. 

While the SOCD mode can be changed with a slider and hotkeys it's current setting can be verified quite easily compared to the input mode. (Except when using a display and maybe some other means I don't know about)

@TheTrainGoes Whenever you're ready let me know. I've been busy as well putting together a prototype version of my controller with LEDs. So I can investigate the issue.

PR #451, opened 2023-08-09
Added display functionality for button profiles

HoboDev:
Exposed wired pin configuration on gamepad
Changed display layouts to use functions with fixed pin layout
Added pin profile indicator
Renamed variables and functions to be more expressive (please forgive me)


bsstephan:
@HoboDev - I don't mind the renames, but I think the introduction of the new pin structures could use so explaining (at least for me).

HoboDev:
The new structures are basically the same as the old single profile structure but with another name. I used activeWire as I didn't come up with a better name.

So button profile 1 = pin assignment = activeWire for the display.
The rest is more or less the same.

Is that what you were looking for?

bsstephan:
> The new structures are basically the same as the old single profile structure but with another name. I used activeWire as I didn't come up with a better name.
> 
> So button profile 1 = pin assignment = activeWire for the display. The rest is more or less the same.
> 
> Is that what you were looking for?

Somewhat, I guess my remaining question is if we need the extra structure at all, or can just refer to profile 1 in the display. Just thinking it might be clearer to cut down on potentially-confusing duplication of the pins, if those two are effectively the same.

bsstephan:
> > The new structures are basically the same as the old single profile structure but with another name. I used activeWire as I didn't come up with a better name.
> > So button profile 1 = pin assignment = activeWire for the display. The rest is more or less the same.
> > Is that what you were looking for?
> 
> Somewhat, I guess my remaining question is if we need the extra structure at all, or can just refer to profile 1 in the display. Just thinking it might be clearer to cut down on potentially-confusing duplication of the pins, if those two are effectively the same.

...and then I guess we probably wouldn't need the `pinProfile` renames either, which would cut down on the diff? Not demanding this, but curious what you think.

HoboDev:
They aren't really necessary. It was just working well in my head that there is a pin assignment and pin profiles.

bsstephan:
> They aren't really necessary. It was just looking good that there is a pin assignment and pin profiles which helped me with the logical separation.

I think we should omit the duplication, then, it's just more to have to mentally manage if there's two things that mean the same thing.

PR #453, opened 2023-08-12
Set the profile number in the config after switching

bsstephan:
I apparently accidentally refactored this away, so adding it back

### DIFF
--- a/src/storagemanager.cpp
+++ b/src/storagemanager.cpp
@@ -143,7 +143,9 @@ PinMappings& Storage::getProfilePinMappings() {
 
 void Storage::setProfile(const uint32_t profileNum)
 {
+	if (profileNum < 1 || profileNum > 4) return;
 	setFunctionalPinMappings(profileNum);
+	this->config.gamepadOptions.profileNumber = profileNum;
 }
 
 void Storage::setFunctionalPinMappings(const uint32_t profileNum)


HoboDev:
May I recommend to move this into this function: ```Storage::setFunctionalPinMappings(const uint32_t profileNum)``` as last line before function end?

This will cover invalid profileNum as they are rejected by this function.
Additionally it is placed in storage where I'd assume saving would happen considering the line is ```Storage::getInstance().setProfile(profileNum);``` suggesting that you set the profile in Storage.

bsstephan:
All of this is a bit funky after I refactored before submitting a PR. 

> May I recommend to move this into this function: `Storage::setFunctionalPinMappings(const uint32_t profileNum)` as last line before function end?

That would also work, but I was thinking setting the pins might logically be different than storing the setting, for some indeterminate light coupling reason.

> Additionally it is placed in storage where I'd assume saving would happen considering the line is `Storage::getInstance().setProfile(profileNum);` suggesting that you set the profile in Storage.

I think I like inside `setProfile` over `setFunctional...`, and the coupling aligns that way, but then as you said I probably need the validity check. I'll take another look at this when I get a chance. Thanks for commenting!

bsstephan:
@HoboDev I moved retaining the setting into `setProfile` with a check for profile num 1-4. Functions as desired on my boards.

PR #456, opened 2023-08-14
Make joystick slider addon configuration akin to SOCD slider's

deeebug:
As per request, this PR makes the joystick slider addon's configuration interface similar to that of the SOCD slider addon.


configs/BentoBox/BoardConfig.h
```diff
@@ -38,7 +38,7 @@
 #define PIN_BUTTON_TURBO -1         // Turbo
 #define PIN_BUTTON_REVERSE -1       // UDLR Reverse
 #define PIN_SLIDER_LS    -1         // Left Stick Slider
```
mthiesen, line 40: You missed this instance of `PIN_SLIDER_LS` during renaming.

mthiesen (changes requested):

configs/BentoBox/BoardConfig.h
deeebug: Oh, this is strange, thank you for catching it! Fixed!

arntsonl (approved):
Neat! This gives our joystick slider more flexibility, I'm into it!

PR #458, opened 2023-08-16
PS add-on callback logic fix

arntsonl:
Had the waiting for callback logic flipped, thanks jfedor!


arntsonl (commented):
Really silly mistake, all fixed

PR #462, opened 2023-08-20
Fixing PIO state machines 

arntsonl:
This fixes our USB host manager start sync, and we can now run PIO state machines on both the PIO PICO USB configuration and the NeoPico PIO code samples.

The reason Core1 worked before is we had to init NeoPico BEFORE we did PIO USB.

### DIFF
--- a/src/usbhostmanager.cpp
+++ b/src/usbhostmanager.cpp
@@ -10,12 +10,9 @@ void USBHostManager::setDataPin(uint8_t inPin) {
 
 void USBHostManager::start() {
     if ( !addons.empty() ) {
-    	// State Machine moved to PIO0:1 and PIO1:1,2 for NeoPico
         pio_usb_configuration_t pio_cfg = PIO_USB_DEFAULT_CONFIG;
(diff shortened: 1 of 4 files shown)


src/usbhostmanager.cpp
```diff
@@ -10,8 +10,12 @@ void USBHostManager::setDataPin(uint8_t inPin) {
 
 void USBHostManager::start() {
     if ( !addons.empty() ) {
+    	// State Machine moved to PIO0:1 and PIO1:1,2 for NeoPico
         pio_usb_configuration_t pio_cfg = PIO_USB_DEFAULT_CONFIG;
         pio_cfg.pin_dp = dataPin;
+        pio_cfg.sm_eop = 2;
+        pio_cfg.sm_rx = 1;
+        pio_cfg.sm_tx = 1;
```
sm00th, line 18: NeoPico utilizes only pio0:0, so reconfiguring tx should be enough.

arntsonl: Yeah, that makes sense. I'll revert back eop and rx

PR #463, opened 2023-08-20
Rename "PS4 Mode" addon

sm00th:
The phrase "PS4 Mode" is used throughout documentation for both the ds4 emulation and the addon that handles authentication. With addition of passthrough authentication it got even more confusing.

This change renames the addon from "PS4 Mode" to "PS4 File Authentication" which better reflects addon's nature.

### DIFF
--- a/docs/ps4.md
+++ b/docs/ps4.md
@@ -20,7 +20,7 @@ The 8-minute timeout works like this:
 5. User unplugs their non-functional controller
 6. Go back to step 1 and repeat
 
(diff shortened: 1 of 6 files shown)


arntsonl:
I don't think I see the need to do this rename, as we've used this nomenclature for 8-9 months.

Technically we're not even emulating PS5, we're running a PS4 Arcade Stick which works on a PS5.

I do like the line in SettingsPage.jsx that says:

"PS4 Mode" add-on if you have the necessary files, or use "PS Passthrough" with a PS4 or PS5 compatible controller/dongle',"

arntsonl:
If you want to update our settings page a bit, I'd resubmit this PR with a "clarification of PS4/PS5 modes". Going to close for now.

PR #470, opened 2023-08-24
Add feature to capture button held pins from board

deeebug:
This PR adds a new functionality to the Pin Mapping configuration screen. It allows the user to press buttons on their controller to assign pins. It can be done for each button individually or all buttons one-by-one.

Currently, as per the implementation, it initializes all valid pins. Within a window of 5 seconds, if the user does nothing, it returns empty. Before anything is returned, all unused pins are also uninitialized.

### DIFF
--- a/www/src/Pages/PinMapping.jsx
+++ b/www/src/Pages/PinMapping.jsx
@@ -1,8 +1,9 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useState, useRef } from 'react';
 import { NavLink } from 'react-router-dom';
 import { Button, Form } from 'react-bootstrap';
 import { AppContext } from '../Contexts/AppContext';
 import Section from '../Components/Section';
+import CaptureButton from '../Components/CaptureButton';
 import WebApi, { baseButtonMappings } from '../Services/WebApi';
 import boards from '../Data/Boards.json';
 import { BUTTONS } from '../Data/Buttons';
@@ -25,8 +26,12 @@ export default function PinMappingPage() {
 	const [buttonMappings, setButtonMappings] = useState(baseButtonMappings);
 	const [selectedController] = useState(import.meta.env.VITE_GP2040_CONTROLLER);
 	const [selectedBoard] = useState(import.meta.env.VITE_GP2040_BOARD);
+	const [capturingButtonIndex, setCapturingButtonIndex] = useState(-1);
+	const [isCapturingPinsInProgress, setIsCapturingPinsInProgress] = useState(false);
+	const controller = useRef();
 	const { buttonLabelType, swapTpShareLabels } = buttonLabels;
 	const { setLoading } = useContext(AppContext);
+	const buttonNames = Object.keys(BUTTONS[buttonLabelType])?.filter((p) => p !== 'label' && p !== 'value');
 
 	const { t } = useTranslation('');
 
@@ -48,12 +53,12 @@ export default function PinMappingPage() {
 		fetchData();
 	}, [setButtonMappings, selectedController]);
 
-	const handlePinChange = (e, prop) => {
+	const handlePinChange = (e, prop, unassignOthersOnConflict) => {
 		const newMappings = { ...buttonMappings };
 		if (e.target.value) newMappings[prop].pin = parseInt(e.target.value);
 		else newMappings[prop].pin = '';
 
-		validateMappings(newMappings);
+		validateMappings(newMappings, unassignOthersOnConflict, prop);
 	};
 
 	const handleSubmit = async (e) => {
@@ -77,7 +82,7 @@ export default function PinMappingPage() {
 		);
 	};
 
-	const validateMappings = (mappings) => {
+	const validateMappings = (mappings, unassignOthersOnConflict, targetButton) => {
 		const buttons = Object.keys(mappings);
 
 		// Create some mapped pin groups for easier error checking
@@ -102,7 +107,6 @@ export default function PinMappingPage() {
 
 		for (let button of buttons) {
 			mappings[button].error = '';
-
 			// Validate required button
 			if (
 				(mappings[button].pin < boards[selectedBoard].minPin ||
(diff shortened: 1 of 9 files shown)


www/src/Pages/PinMapping.jsx
```diff
@@ -236,7 +271,12 @@ export default function PinMappingPage() {
 												max={boards[selectedBoard].maxPin}
 												isInvalid={buttonMappings[button].error}
 												onChange={(e) => handlePinChange(e, button)}
+												onFocus={() => setSelectedButton(button)}
```
Pelsin, line 274: Unsure if this is used

www/src/Pages/PinMapping.jsx
Pelsin, line 1: Don't forget to run `npm run format` or some automatic formatter, if you use one and it's not working it might be on me ;D

Pelsin:
So awesome, love this functionality!

I have some structural opinions that *might* help make this more reusable.
That or i'm missing something :)

It feels somewhat redundant that we have `handlePinChange` and `getHeldPins`.
We should be able to separate the recording to a hook or a button, i made a quick example here for the specific inputs(not for all) but i do believe we can use it.

This example also shows how to solve a bug that we have now. The request keeps going even if you change page, leaving it hanging.
Also happens when pressing `Capture all button pins`

(i will send video on discord as the file size is to big)

The only thing that is needed for this to work on the pinMapping page is.
```javascript
<CaptureButton
  onChange={(pin) =>
    handlePinChange({ target: { value: pin } }, button)}
/>
```
With this we should also be able to use it on other pages.

Here is the captureButton
```javascript
import axios from 'axios';
import React, { useEffect, useRef, useState } from 'react';
import { Button, Modal } from 'react-bootstrap';
import { useTranslation } from 'react-i18next';

const CaptureButton = ({ onChange }) => {
	const { t } = useTranslation('');
	const controller = useRef<AbortController>();
	const [modalVisible, setModalVisible] = useState(false);

	const newAbortSignal = () => {
		controller.current = new AbortController();
		return controller.current.signal;
	};

	const getHeldPins = async () => {
		setModalVisible(true);
		const { data } = await axios
			.get('http://localhost:8080/api/getHeldPins', {
				signal: newAbortSignal(),
			})
			.then(({ data }) => {
				onChange(data?.heldPins?.at(0));
				setModalVisible(false);
			});

		onChange(data?.heldPins?.at(0));
		setModalVisible(false);
	};

	useEffect(() => () => controller?.current?.abort(), []);

	const closeModal = () => {
		controller?.current?.abort();
		setModalVisible(false);
	};

	return (
		<>
			<Modal show={modalVisible} onHide={() => closeModal()}>
				<Modal.Header closeButton>
					<Modal.Title>Waiting for button press</Modal.Title>
				</Modal.Header>
				<Modal.Body>
					<div className="d-flex justify-content-center align-items-center loading">
						<div className="spinner-border" />
					</div>
				</Modal.Body>
				<Modal.Footer>
					<Button variant="danger" onClick={() => closeModal()}>
						Cancel
					</Button>
				</Modal.Footer>
			</Modal>
			<Button
				className="ms-3"
				variant="secondary"
				onClick={() => getHeldPins()}
			>
				{t('PinMapping:capture-button-label')}
			</Button>
		</>
	);
};

export default CaptureButton;

```

Pelsin:
Also! My suggestion can be a future improvement, no need to do it in this PR if it's out of scope

deeebug:
@Pelsin thank you very much!
These are great suggestions, I'll try to get them integrated.

www/src/Pages/PinMapping.jsx
deeebug: Oh this is not!
It was for something else that I was planning and it didn't work out. This will be cleaned.

deeebug: Cleaned up

InfraredAces:
Testing with a FlatboxRev5RGB

The individual buttons works exactly as intended. Loaded incorrect firmware to controller and after navigating to the Web Configurator, performed the following steps.

1. Press button on screen
2. Prompted for input
3. Press button on controller
4. Pin is mapped on controller
5. Repeated steps 1-4 until all buttons were mapped
6. Once save button on page is pressed and controller restarted, the controller works as intended.

image

## Issue

When using the `Assign Gamepad Pins` button in the `Configuration>Pin Mapping` page, initially it will register the negative edge of button presses as additional button presses. This means that the process is incremented by 2 steps when only 1 is met.

After a period of time leaving the `Assign Gamepad Pins` feature will proceed, but the mapping process stops registering inputs consistently. It will sometimes read inputs and others, it will time out until the next input comes up. Once this occurs, I am unable to navigate to any other pages in WebConfig, nor does refreshing the page resolve the issue.

## Recommended Changes

The labels used in the modal dialog for mapping uses GP2040-CE labels regardless of what the Label Dropdown is set to. It would be helpful to have the labels match with the user selection

deeebug:
@InfraredAces thank you very much for testing and a very elaborate report!
Turns out we needed a debouncer for the pins.

Also the labels have been corrected.

InfraredAces:
As discussed in the Discord, on RP2040 boards that have access to Pin 29, this feature does not work when assigning buttons to Pin 29 and require manually entering the pin for assignment

Otherwise I was not able to find issue with the PR. All other pins are assignable by pressing the button, both individually and when entering then sequentially using the `Assign Gamepad Pins` button in WebConfig.

arntsonl:
That is really interesting, and makes me think about something maybe we should add to BoardConfig.h

Would it make sense to add an array definition, something like

#define BOARD_AVAILABLE_PINS 0,1,2,3,4,5...24,25,26,27,28,29

then we can just wrap that in a static const uint32_t boardPins[] = {BOARD_AVAILABLE_PINS}; ?

For this PR, I'm going to merge it since its working as expected. Then I'll make a new issue to add board pin definitions to BoardConfig.h

deeebug:
That sounds good to me!
Thank you for noticing the issue in WebApi.js, this has been reverted.

arntsonl (approved):
Looks great!

PR #472, opened 2023-08-27
GPIO labels

mindplay-dk:
Clarify the fact that these are logical GPIO numbers, as opposed to physical pin numbers - since actual pin numbers are device-dependent, and since there is a risk of someone ignorant (like me!) thinking these are *actual* pin numbers, it's probably safer to designate them as "GPIO" pins.

### DIFF
--- a/www/src/Locales/en/PinMapping.jsx
+++ b/www/src/Locales/en/PinMapping.jsx
@@ -1,5 +1,5 @@
 export default {
-	'header-text': 'GPIO Mapping',
+	'header-text': 'GPIO Pin Mapping',
 	'sub-header-text':
 		'Use the form below to reconfigure your button-to-GPIO pin mapping.',
 	'alert-text':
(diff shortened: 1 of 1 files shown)


arntsonl:
Hey there @mindplay-dk , we're going to stick with the "pin" language in our code base. Thank you for the PR suggestion, please stop by our discord at https://discord.gg/Nc2ByRMk if you'd like to chat with the dev group about changes and suggestions. We're very active in #firmware-development !

mindplay-dk:
Perhaps your documentation should clarify this then?

The software itself adapts in other ways to the device it's installed on - some people might expect "pin" to be the actual pin numbers on the device. Well, why not. The printed pin numbers on the Pico are "1, 2, 3", exactly the way they're shown in the manual *and* in the software - while the printed GPIO pins are labeled "GP1, GP2", etc.

I'm obviously new at this. Just trying to help others find their way and avoid mistakes. 🙂

Issue #431, opened 2023-07-28
Issue - Chase animation bug when increasing speed


InfraredAces:
### Prerequisites

Please check the following before posting an issue / bug report.

- [ X ] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [ X ] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [ X ] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [ X ] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: [0.7.3](https://github.com/OpenStickCommunity/GP2040-CE/releases/download/v0.7.3/GP2040-CE_0.7.3_FlatboxRev5RGB.uf2)
* Name of device: Flatbox Rev5 RGB
* Link to where you purchased the device: [OpenStickCommunity](https://discord.com/channels/1049366310389289001/1049371306958782495/1125457956775931956)
* Is this a custom build?: No
* Operating System: 6.4.1-arch2-1 (64-bit)
* Browser: Chrome

### Expected Behavior

Using the LED chase animation, repeatedly press S1 + S2 + R2, eventually reach a maximum speed where LEDs chase.

### Current Behavior

Using the LED chase animation, repeatedly press S1 + S2 + R2, eventually, the LEDs stop chasing and when pressed the Pressed LED color stays without disappearing.

Changing the LED animation works, but when going back to chase animation, above behavior continues.

Lowering the animation speed allows the chase animation to work once more, but only after changing LED animations away and then back to the chase animation.

### Steps to Reproduce

1. Plug in controller
2. Change the LED animation to Chase
3. From minimum speed, press S1 + S2 + R2 (LED Parameter Down) 32 times (will still happen if more than 32)
4. LED stop chasing
5. Press buttons and Pressed LED color will stay
6. Change the LED animation to another animation
7. Change the LED animation back to Chase
8. LED does not chase and pressed button LED color will remain
9. Lower chase animation speed by pressing S1 + S2 + R1 a few times
10. Change the LED animation to another animation
11. Change the LED animation back to Chase
12. LED does begin chasing again and pressed button LED color will no longer remain

### Screenshots & Files

[...]

PR #474, opened 2023-08-28
Small fixes to LED Animation system

Procdox:
Fix for https://github.com/OpenStickCommunity/GP2040-CE/issues/431, animation state breaks for cycling animations when they are made too fast. 
Also includes a fix to a potential memory corruption via Animation pointers in the AnimationStation.
Another small change to the RBG struct constructors to make them Trivial classes.

Please note, as my hitbox has not yet arrived I am unable to test these changes past simply ensuring they compile.

### DIFF
--- a/lib/AnimationStation/src/Animation.hpp
+++ b/lib/AnimationStation/src/Animation.hpp
@@ -10,7 +10,11 @@
 #include "NeoPico.hpp"
 
 struct RGB {
-  RGB() : r(0), g(0), b(0) {}
+  // defaults allows trivial constructor, avoiding compiler complaints and avoiding unnessecary initialization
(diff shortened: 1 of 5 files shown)


InfraredAces:
Testing with FlatboxRev5RGB

This issue and behavior stated in #431 are still present. This is the case even after flash_nuking the board and then flashing the build firmware from the workflow run for this PR.

arntsonl (approved):
Should prevent crashes on animation change! Thanks!

Procdox:
I was able to consistently reproduce the issue before my changes. I was not able to do so with my changes.
However, @InfraredAces reports that they are still seeing the same or a similar issue post changes. So #431 requires further investigation.

PR #478, opened 2023-08-29
Rev5 Southpaw Config 

SkylaHila:
Created southpaw config and tested on my southpaw controller, as well as sent it to someone who built a southpaw and they confirmed it working.


arntsonl (approved):
Looks good!

PR #481, opened 2023-08-29
Fix hotkeys with input addons

deeebug:
This is a simple fix that allows input addons to use the hotkeys feature.
This was tested on the Keyboard Host addon and is to be noted that the boot selection (selection of controller types/web-config) didn't work due to the delay in USB initialization which might not be present in other addons.

Please test and merge, thanks!

### DIFF
--- a/src/gp2040.cpp
+++ b/src/gp2040.cpp
@@ -159,11 +159,11 @@ void GP2040::run() {
 	#if GAMEPAD_DEBOUNCE_MILLIS > 0
 		gamepad->debounce();
 	#endif
-		gamepad->hotkey(); 	// check for MPGS hotkeys
-		rebootHotkeys.process(gamepad, configMode);
-
 		// Pre-Process add-ons for MPGS
 		addons.PreprocessAddons(ADDON_PROCESS::CORE0_INPUT);
+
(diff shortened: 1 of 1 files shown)


InfraredAces:
Testing this PR with the Keyboard Host Add-on, the following Hotkeys function properly

- No Action
- DPad Digital
- DPad Left Analog
- DPad Right Analog
- SOCD Up Priority
- SOCD Neutral
- SOCD Last Win
- SOCD First Wins
- Invert X Axis
- Toggle 4-Way Joystick Mode
- L3 Button
- R3 Button
- Touchpad Button
- Restart GP2040-CE

The following hotkeys do not function properly

- Home Button (when triggering hotkey, S1 is triggered instead)
- Capture Button (when triggering hotkey, A1 is triggered instead)
- Invert Y Axis (does not invert Y axis)

Unable to verify the following hotkeys

- SOCD Cleaning Off
- Toggle DDI 4-Way Joystick Mode
- Load Profile # 1-4

arntsonl:
> 

We can file these specific hotkeys as an issue/bug as they are unrelated to @deeebug 's change. Will be a good PR to figure out what is going on with those functions

arntsonl (approved):
Makes sense, slight ordering change!

PR #485, opened 2023-08-31
Remove PS4 counter

jfedor2:
Gets rid of the counter in the PS4 input report.

It doesn't do anything except add 1ms of latency.


arntsonl:
The counter is part of the DS4 packet, which we want to keep the same as the real DS4 report. This is a to-investigate, we need to compare against other dongles and third-party controllers.

arntsonl (approved):
I'm going to make an issue to investigate this further, but for now we can remove.

dogtopus:
This may be needed once we have the IMU support, but we need more tests on this.

PR #489, opened 2023-09-01
Extract addons

Pelsin:
Separating addons from addon page as way to decouple some logic.
Found some errors and typos, also changed some styling for warning and disclamer

Skärmavbild 2023-09-01 kl  19 03 13

Skärmavbild 2023-09-01 kl  19 03 07

There was some addons that had no initial state and some missing some validation, i think SNES addon still is missing some. But hopefully this changes helps working with addons in the future.

Please help test if possible, as every addon has been touched.


arntsonl (approved):
Very clean!

PR #491, opened 2023-09-02
setHasFlags() correctly handles PB repeated fields

mthiesen:
This PR fixes a problem with `setHasFlags()` discovered by @deeebug. Previously, the function did not correctly handle Protobuf repeated fields containing sub-messages. With this fix, the workaround of manually setting `has_XXX` flags is no longer necessary.


arntsonl (approved):
Makes sense, fixes our has_xxxxx = issue in protobuf

PR #494, opened 2023-09-04
Dev board mode

Pelsin:
Got choice paralysis from all the different ways of solving this.

Maybe just setting a env in the script is more straightforward as @deeebug said, i'll add cross-env if you don't object, there was a post a few years ago declaring the lib feature complete, hence it being archived: https://github.com/kentcdodds/cross-env/issues/257

Setting PR as draft, don't have the possible to test against a board ATM. I'm guessing we need to fix CORS issues.


arntsonl (approved):
Let's get it in!

PR #495, opened 2023-09-04
Remove BB12C references and inline remaining usage

Procdox:
This branch seeks to remove unused portions of BBI2C and direct references to it to remove ambiguity.
BBI2C was replaced with inlinable I2C_IF interface and unused/non-functional discover and bwire functionality were removed.

Tested with a SSD1306 and stock Pico on 9/4.


mikepparks:
I'll put my $0.02 in for the rest of the team's consideration.

While we don't necessarily seem to have any sort of _hard_ code style conventions defined, I would likely say that we may want to keep the camelCase naming for consistency's sake.  Also, using .hpp for everything does seem to deviate from the .cpp/.h separation found in other libraries already in use.  I would also say that if this is a rewrite of the already heavy modification to BBI2C, which was effectively in-name only, then we don't necessarily need to make reference to it in documentation since it's no longer in use.  That said, a very good first effort.

arntsonl:
Let's keep our headers .h rather than .hpp

arntsonl:
We're going to be rewriting I2C instead of patching out add-ons. Closing this draft as we don't want to rewrite a rewrite.

Issue #498, opened 2023-09-05
Enabling Reverse function prevents the direction inputs from pin 16-19 from getting recognized


jchen2601:
1 - What device are you using?
TheTrain RP2040 Advanced Breakout Board version 5.2E
2 - Where did you buy it? (post link to sales page)
TheTrain from /r/fightsticks Discord
3 - What version of the firmware are you currently running? (check this via web-config)
v0.7.4
4 - Which distribution of the firmware are you currently running? (copy and paste the name of the firmware file you loaded on)
GP2040-CE_0.7.4_RP2040AdvancedBreakoutBoard.uf2
5 - What is the problem that you are having?
Enabling Reverse function to button on pin 23 disables all the other directional buttons that I bound to pin 16-19. How do I enable Reverse without disabling the other directional buttons?
6 - What trouble shooting steps have you already done?

Disabling Reverse allows the pin 16-19 directional buttons to be used again. I can bind additional regular directional input UDLR to the pin 23 button no problem. When I reenabled Reverse to pin 23, the pin 16-19 directional buttons get disabled again. Enabling Reverse for Left and Down input while disabling Up and Down input Reverse still disables the pin 16-19 directional buttons. Enabling Reverse for pin 29 also produces the same issue. I am connecting with Xinput. 

![image (2)](https://github.com/OpenStickCommunity/GP2040-CE/assets/144094861/3987e914-fb04-4d83-a226-0f00b6ae0a1c)

![20230904_165949](https://github.com/OpenStickCommunity/GP2040-CE/assets/144094861/87cf7b85-7ed0-42b9-9738-d303b4cdddd9)

TheTrainGoes:
This has been fixed and merged into main via #499 .

Closing issue as resolved.

jchen2601:
How do I apply this fix? I downloaded the firmware for RP2040 Advanced Breakout from the GP2040-CE website and applied it to TheTrain breakout board. Enabling Reverse on pin 29 still disables directional buttons on pin 16-19. I can't assign directional button to other pins because they are connected via the 20pin harness or to the 16-19 pin of the screw terminal that corresponds to directions.

TheTrainGoes:
We have not put this to release yet.  It's been merged to main so you can compile yourself if you like otherwise you will need to wait until our next release is up.

jchen2601:
> We have not put this to release yet. It's been merged to main so you can compile yourself if you like otherwise you will need to wait until our next release is up.

[...]

PR #499, opened 2023-09-05
Truncated mask broke high pin dpad inputs

Procdox:
Fix for [#498](https://github.com/OpenStickCommunity/GP2040-CE/issues/498)
Reverse Inputs addon would break all dpad inputs if they were assigned to high pins ( pin 8, mask >= 256 ).
This was due to truncation of the mask in the ReverseInput::input parameters.

Recreated and fixed with standard pico:
Reverse pin 22
DPad pins: 6 7 8 9
This caused 8 and 9 to not function before the fix.

### DIFF
--- a/headers/addons/reverse.h
+++ b/headers/addons/reverse.h
@@ -25,7 +25,7 @@ public:
     virtual std::string name() { return ReverseName; }
 private:
     void update();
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Should be fine

PR #500, opened 2023-09-05
feat: Enable Tilt Value Configuration via WebConfig

smashBrosKanu:
### Overview
This pull request adds the ability to configure the Tilt value through WebConfig.

### Background
Different characters may require unique Tilt values to perform their intended actions. Given this, it is desirable to allow users to set these values themselves.

### Usage
Users can now freely set the Tilt value through the WebConfig settings interface.

### Impact
This change affects the WebConfig UI and the logic that processes the Tilt value.

### Testing
Manual testing was conducted in a local environment to ensure that the Tilt value could be accurately set.

Please review. Thank you!


TheTrainGoes:
Looks neat!  I will check this out today when I have some time

www/src/Locales/zh-CN/AddonsConfig.jsx
```diff
@@ -75,7 +75,15 @@ export default {
...
 	'tilt-1-pin-label': 'Tilt 1 引脚',
+	'tilt-1-factor-left-x-label': 'Tilt 1 Factor Left X',
+	'tilt-1-factor-left-y-label': 'Tilt 1 Factor Left Y',
+	'tilt-1-factor-right-x-label': 'Tilt 1 Factor Right X',
+	'tilt-1-factor-right-y-label': 'Tilt 1 Factor Right Y',
 	'tilt-2-pin-label': 'Tilt 2 引脚',
+	'tilt-2-factor-left-x-label': 'Tilt 2 Factor Left X',
+	'tilt-2-factor-left-y-label': 'Tilt 2 Factor Left Y',
+	'tilt-2-factor-right-x-label': 'Tilt 2 Factor Right X',
+	'tilt-2-factor-right-y-label': 'Tilt 2 Factor Right Y',
```
Pelsin, line 86: Same here

proto/config.proto
```diff
@@ -372,18 +372,26 @@ message TiltOptions
...
-	optional int32 tiltLeftAnalogDownPin = 6;
-	optional int32 tiltLeftAnalogLeftPin = 7;
-	optional int32 tiltLeftAnalogRightPin = 8;
-	optional int32 tiltRightAnalogUpPin = 9;
-	optional int32 tiltRightAnalogDownPin = 10;
-	optional int32 tiltRightAnalogLeftPin = 11;
-	optional int32 tiltRightAnalogRightPin = 12;
-
-	optional SOCDMode tiltSOCDMode = 13;
+	optional int32 factorTilt1LeftX =  3;
```
mthiesen, line 375: You should never change the tag numbers of existing protobuf fields, this breaks backwards compatibility. You can add as many new fields as you like, and you can change the order of the fields in a struct as long as the tag numbers of existing fields do not change.

mthiesen (changes requested):

arntsonl (approved):
Looks good!

PR #501, opened 2023-09-06
Wii addon refactor and button/analog mapping support

mikepparks:
This also includes support for mapping the following extension controllers:

* Guitars (GH/GHWT)
* Drums (GHWT/BH)
* Taiko drum (TnT)
* Turntables (DJH)


lib/WiiExtension/WiiExtension.h
```diff
@@ -52,6 +58,10 @@
 #define WII_EXTENSION_DEBUG false
 #endif
 
+#ifndef WII_EXTENSION_ENCRYPTION
```
arntsonl, line 61: If these don't do anything yet, might just put a comment above saying like

// NO IMPLEMENTATION YET

or something in case someone else picked up this code

lib/WiiExtension/WiiExtension.h
```diff
@@ -61,12 +71,16 @@
 #endif
 
 #ifndef WII_EXTENSION_CALIBRATION
```
arntsonl, line 73: Same with this. If we wanted to enable/disable extension calibration, we should make a web config checkbox. Otherwise we can assume we always want it :)

lib/WiiExtension/WiiExtension.h
```diff
@@ -145,59 +114,17 @@ class WiiExtension {
...
-    uint16_t _triggerPrecision1From = WII_ANALOG_PRECISION_0;
-    uint16_t _triggerPrecision1To = WII_ANALOG_PRECISION_0;
-    uint16_t _triggerPrecision2From = WII_ANALOG_PRECISION_0;
-    uint16_t _triggerPrecision2To = WII_ANALOG_PRECISION_0;
-
-    uint8_t _guitarType   = WII_GUITAR_UNSET;
-
-    uint16_t map(uint16_t x, uint16_t in_min, uint16_t in_max, uint16_t out_min, uint16_t out_max);
-    uint16_t calibrate(uint16_t pos, uint16_t min, uint16_t max, uint16_t center);
+    uint8_t _lastRead[16] = {0xFF};
```
arntsonl, line 127: Looks like we could get rid of this since its commented out in all of your debugs?

lib/WiiExtension/extensions/ClassicExtension.cpp
```diff
@@ -0,0 +1,210 @@
...
+        buttons[WiiButtons::WII_BUTTON_A]                      = !((inputData[7] & 0x10) >> 4);
+        buttons[WiiButtons::WII_BUTTON_X]                      = !((inputData[7] & 0x08) >> 3);
+        buttons[WiiButtons::WII_BUTTON_ZR]                     = !((inputData[7] & 0x04) >> 2);
+        buttons[WiiButtons::WII_BUTTON_LEFT]                   = !((inputData[7] & 0x02) >> 1);
+        buttons[WiiButtons::WII_BUTTON_UP]                     = !((inputData[7] & 0x01) >> 0);
+    } else {
+        // unknown
+    }
+
+#if WII_EXTENSION_DEBUG==true
```
arntsonl, line 203: Just for cleanliness, we'll probably want to remove this before release if you're not using it. Okay to keep it into this refactor!

proto/config.proto
```diff
@@ -433,11 +433,121 @@ message PSPassthroughOptions
...
+
+    message StickOptions
+    {
+        optional AnalogAxis x = 1;
+        optional AnalogAxis y = 2;
+    }
+    
+	message NunchukOptions
+	{
+		optional int32 buttonC = 1;
```
arntsonl, line 451: for some reason, some of the indents are tabs and some are spaces

src/addons/wiiext.cpp
```diff
@@ -30,190 +32,501 @@ void WiiExtensionInput::setup() {
...
+                    break;
+            }
+        }       
+    }
 }
 
-uint16_t WiiExtensionInput::map(uint16_t x, uint16_t in_min, uint16_t in_max, uint16_t out_min, uint16_t out_max) {
-    return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
+uint16_t WiiExtensionInput::getAverage(std::vector<WiiAnalogChange> const& changes) {
+    std::vector<uint16_t> values;
```
arntsonl, line 501: we could do this without a vector I think?
```
    uint16_t values = 0;
    for (auto currVal = changes.begin(); currVal != changes.end(); ++currVal) {
        values += currVal->analogValue;
    }
    return values / changes.size();

```

I think?

src/addons/wiiext.cpp
```diff
@@ -30,190 +32,501 @@ void WiiExtensionInput::setup() {
...
+
+    for (auto currVal = changes.begin(); currVal != changes.end(); ++currVal) {
+        values.push_back(currVal->analogValue);
+    }
+
+    return std::accumulate(values.begin(), values.end(), 0) / values.size();
 }
+
+uint16_t WiiExtensionInput::getDelta(std::vector<uint16_t> const& changes, uint16_t baseValue) {
+    uint16_t value = baseValue;
```
arntsonl, line 515: Yeah getAverage could do this I think?

arntsonl (changes requested):
Looks good! Small changes for styling but its a great refactor

src/addons/wiiext.cpp
mikepparks: getAverage is for smoothing out values, where getDelta is for knowing how much of a change needs to be applied to a given point. It adjusts to account for neutral directions resetting the analog state if it changed earlier in the loop.

Example: LS and RS are mapped to L analog output.  LS gets read first, then RS. LS moves, RS stays neutral. 
 Without getDelta, LS adjusts L.X to xFFFF, but RS.X is 7FFF, so L.X stays x7FFF, undoing the change.  With getDelta, the previous change gets adjusted by later changes before sending them out.  This could use some tuning, but it was to prevent later analog neutral states from dropping earlier changes.

src/addons/wiiext.cpp
mikepparks: You're right. I'll make this simpler.

proto/config.proto
mikepparks: Will be fixed in next push.

lib/WiiExtension/WiiExtension.h
mikepparks: Will put this behind WII_EXTENSION_DEBUG

lib/WiiExtension/WiiExtension.h
mikepparks: I want to leave this as is since not all devices currently support on-device calibration.  The fallback should be enough for _most_ cases, but we may need further testing on third party to know for certain.

lib/WiiExtension/WiiExtension.h
mikepparks: Something as food for thought is that I'm also going to be breaking this library out as standalone for RP2040 usage, so a lot of these would make more sense being included in a generic sense.  I'll make some commentary on it however.

www/src/Addons/Wii.tsx
```diff
@@ -43,91 +47,280 @@ export const wiiState = {
...
+        let wiiMap = wiiButtonsMap
+            .reduce(
+                (o, i) => Object.assign(o, { [i.getAttribute('controlid')+'.button'+i.getAttribute('buttonid')]: parseInt(i.value) }), {});
+
+        let wiiAnalogMap = Array.from(document.querySelectorAll('#WiiExtensionAddonOptions .wii-analogs'));
+        wiiMap = wiiAnalogMap
+        .reduce(
+            (o, i) => {
+                let modeID = i.value;
+                let joyMode = WII_JOYSTICK_MODES.filter((o,i) => o.value == modeID)[0];
```
Pelsin, line 106: Filter is usually used for retrieving one or more based on the predicate.
For one item `Array.find` is usually the way.
Since you only need one here we can simplify it a tiny bit and make it reusable
```javascript
const getJoystickMode = (searchValue: number) =>
	WII_JOYSTICK_MODES.find(({ value }) => value === searchValue);
```

www/src/Addons/Wii.tsx
```diff
@@ -43,91 +47,280 @@ export const wiiState = {
...
+                </Row>
+                <Row className="mb-3">
+                    <Tabs
+                        defaultActiveKey={`wii${WII_EXTENSION_CONTROLS[0].id}Config`}
+                        id="wiiControllerConfig"
+                        className="mb-3"
+                        fill
+                    >
+                        {WII_EXTENSION_CONTROLS.map((controlObj, controlID) => (
+                            <Tab eventKey={`wii${controlObj.id}Config`} title={t(`WiiAddon:controller-${controlObj.id.toLowerCase()}`)}>
```
Pelsin, line 264: We are getting some warnings in the console:
`Warning: Each child in a list should have a unique "key" prop.` 

We have these in other places but if possible add a unique `key` prop when looping some components

www/src/Addons/Wii.tsx
```diff
@@ -43,91 +47,280 @@ export const wiiState = {
 	wiiExtensionSpeed: 400000,
 };
 
+const WII_EXTENSION_CONTROLS = [
```
Pelsin, line 50: I've added ways for us to normalize formating, try running ´npm run format´
Some have reported it not really working, but worth trying!

www/src/Addons/Wii.tsx
```diff
@@ -43,91 +47,280 @@ export const wiiState = {
...
+                            <Tab eventKey={`wii${controlObj.id}Config`} title={t(`WiiAddon:controller-${controlObj.id.toLowerCase()}`)}>
+                                <Row className="mb-3">
+                                    {controlObj.inputs.digital?.length > 0 ? <h1>{t('WiiAddon:section-digital')}</h1> : ''}
+                                    {controlObj.inputs.digital?.map((buttonObj,buttonID) => (
+                                        <div className="col-sm-12 col-md-6 col-lg-2 mb-2" key={`wiiExtensionController${controlObj.id}Digital${buttonID}`}>
+                                            <label className="form-label" htmlFor={`wiiExtensionController${controlObj.id}Button${buttonObj.id}`}>{t(`WiiAddon:controller-button-${buttonObj.id.toLowerCase()}`)}</label>
+                                            <select className="form-select-sm form-control wii-buttons" controlid={`${controlObj.id.toLowerCase()}`} buttonid={`${buttonObj.id}`} id={`wiiExtensionController${controlObj.id}Button${buttonObj.id}`} value={wiiControls[controlObj.id.toLowerCase()+'.button'+buttonObj.id]} onChange={(e) => setWiiControlEntry(controlObj.id, buttonObj.id, e)}>
```
Pelsin, line 270: We can remove `setWiiControlEntry` function by changing the onChange:
```javascript
onChange={(e) =>
  setWiiControls((controls) => ({
    ...controls,
    [`${controlObj.id.toLowerCase()}.button${controlObj.id}`]: e.target.value,
  }))
}}
```

www/src/Addons/Wii.tsx
mikepparks: This worked nicely with some adjustments for number type.

www/src/Addons/Wii.tsx
mikepparks: I would need to break down react-bootstrap's Tab to a NavItem and such to be able to apply a key here.  Would take some doing, but that seems to be what's going on here.

image

www/src/Addons/Wii.tsx
mikepparks: I'd have to go through this manually because it breaks up formatting in a way that isn't very easy on my eyes.

www/src/Addons/Wii.tsx
mikepparks: Fixed this with a fresh night of sleep. Sometimes it's not very clear where the warnings come from in the console.

arntsonl (approved):
Looks great, thank you for updating things!!

PR #504, opened 2023-09-06
convert core GPIO pin mappings from button-to-pin to pin-to-action

bsstephan:
this changes the core board config and gamepad behavior to be that the settings/user configures what each pin does, rather than the old behavior which configures what pin each button is assigned to. this allows for the new configuration to have one canonical location for what each of the pins does, and removes the limitation that a button can only normally be activated by one pin --- now it is trivial to assign many pins to take the same action

"actions" are all button presses right now, but they may be other things in the future as this branch expands. this is just the core functionality and migration code for now, further changes (removing extra input addon, refactoring DDI, maybe refactoring the sliders, working the profiles into this scheme, etc.) to come


arntsonl (commented):
So clean!!!

arntsonl (approved):
Looks great to me!

PR #507, opened 2023-09-07
Add flatbox rev 5 southpaw firmware to downloads page on website

SkylaHila:
Everything should work as intended once the firmware for the southpaw is added to the releases tab


arntsonl:
@SkylaHila we don't add download links until we do a release.

So the Southpaw config will go in the 0.7.5 release when we do that, and will be available on our website after that happens. We also have to update a few other places including the assets, so we will handle this on the release side when that happens.

PR #508, opened 2023-09-07
Add visible board config info with targeted download URL

deeebug:
The following updates add visible board config label and image name to the web-config.
The download URL has also been updated. It now points to the board specific image file, if available.

This should help users identify their boards and make upgrades easier. Also, potentially prevent issues caused by downloading wrong images (as I have).

Please check and merge, thanks!


arntsonl (approved):
Neat! Bakes the board name into the protobuf, will be much easier to figure out whats going on in certain web configs

PR #509, opened 2023-09-07
Added a redirect to uf2 based on url

Pelsin:
https://gp2040-ce.info//#/download?RP2040AdvancedBreakoutBoard

Not the cleanest solution as the user will end up on the download page before the uf2 gets download, might be good enough for now.

The board name in the search param is case insensitive


arntsonl (approved):
Looks good! Matches what debug put in

PR #511, opened 2023-09-08
add scaled radial deadzones

NickGuyver:
Replace the axial deadzone calculations with scaled radial deadzone as demonstrated here - [https://web.archive.org/web/20210827143735/https://www.gamedeveloper.com/disciplines/doing-thumbstick-dead-zones-right](https://web.archive.org/web/20210827143735/https://www.gamedeveloper.com/disciplines/doing-thumbstick-dead-zones-right)

There are a few calculations shared between this and the adjustCircularity function so that was brought into process.


arntsonl (approved):
Yeah its a complicated fix, but we can patch as needed!

PR #512, opened 2023-09-08
Bluetooth Prototype

Procdox:
Proof of concept and a lot of the BTStack / Bluetooth stuff figured out.
Still needs a lot of work, so please feel free to ask questions and contribute.

Check lib/BTAdapter/readme.md for more details

Do not download the artifacts and expect this to work perfectly, I cannot stress enough that this is a prototype.


arntsonl:
Super cool proof of concept! I think we'll need a lot of work to get this integrated into our project, but you really showed its possible.

This would be a SUPER good candidate to branch off into a GP2040-BT project or something similar if we need to rip out a lot of USB code. If we handle it with like a slider or something like 8bitdo does, that might be a better way to integrate it into gp2040-ce. Or add some code if TinyUSB connects then disable bluetooth support.

Please feel free to resubmit the PR when its closer to a stage where you want us to go through the code and review it. Thank you!

PR #514, opened 2023-09-11
Full sleep removal from core0 and core1 loops

arntsonl:
Need to do more testing on this, but this matches what @mthiesen originally thought about removing sleeps. I think the missing piece was not clearing feature data and requesting a "get report" if the endpoints are busy.

I need to test this more, so this is a draft copy


arntsonl:
I think this is good, I was testing it with my other changes and didn't have any side effects.

PR #524, opened 2023-09-15
Adds search functionality to https://gp2040-ce.info/

Pelsin:
Skärmavbild 2023-09-15 kl  12 09 10

**Could not find a way to search for dynamically generated content in "Downloads"**
Before merging it might be worth considering if this caveat is a deal-breaker.


arntsonl (approved):
Nice!!

PR #526, opened 2023-09-16
Add option to turn off RGB & OLED on host suspend

deeebug:
This adds an option to have the RGB LEDs, Player LEDs and the OLED Display turned off when the host (PC, console, etc.) goes to sleep.

This hasn't been tested on a console, but should function as per the USB protocol.

### DIFF
--- a/headers/addons/i2cdisplay.h
+++ b/headers/addons/i2cdisplay.h
@@ -62,6 +62,10 @@
 #define SPLASH_DURATION 7500 // Duration in milliseconds
 #endif
 
+#ifndef DISPLAY_TURN_OFF_WHEN_SUSPENDED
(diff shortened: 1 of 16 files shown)


headers/addons/i2cdisplay.h
```diff
@@ -62,6 +62,10 @@
 #define SPLASH_DURATION 7500 // Duration in milliseconds
 #endif
 
+#ifndef DISPLAY_TURN_OFF_WHEN_SUSPENDED
```
ayan4m1, line 65: Nitpick, and it would be a significant refactor, but I like `DISPLAY_DISABLE_WHEN_SUSPENDED` a bit more

arntsonl: That's pretty nitpicky, going to resolve this one

PR #528, opened 2023-09-17
Fix builds on arm64

ayan4m1:
Running the "linux" version of makefsdata on a Raspberry Pi or similar arm64 SBC will fail with

> -bash: ./makefsdata: cannot execute binary file: Exec format error

As expected, since the architectures are incompatible. This is not addressed by the [documentation](https://gp2040-ce.info/#/development?id=linux-setup-raspberry-pi) on the Pi.

This PR adds an arm64 binary for makefsdata and adds a tiny bit of logic to makefsdata.js which will detect and append the architecture to the binary path before running it.

Also, if an architecture is missing, the error will point the user to the fact that their specific architecture does not have a supported binary, which hopefully aids self-debugging.

I built the arm64 binary on a Raspberry Pi 4 with Debian 12 and the required patches mentioned in the linux_compile_makefsdata script.


arntsonl:
This will break functionality for our other builds, and is specific to a very one-off build environment.

ayan4m1:
I'd be happy to address any regressions if you can point out what is breaking. There is documentation on building on a Raspberry Pi which leads people to believe it is a supported dev environment.

PR #529, opened 2023-09-18
Reset timer in Turbo extension when turbo button is pressed

sirrow:
This is for avoiding ignoring beginning of input if the input began during bTurboFlicker==true.
In order to avoid the ignoring, this change make bTurboFlicker=false and reset timer for inverting bTurboBlicker in case no turbo button was pressed at last sampling and some of turbo button is pressed at the time.

### DIFF
--- a/headers/addons/turbo.h
+++ b/headers/addons/turbo.h
@@ -83,6 +83,11 @@
 #define PIN_SHMUP_DIAL -1
 #endif
 
+// RESET TIMER
+#ifndef RESET_TIMER_ENABLED
+#define RESET_TIMER_ENABLED 0
+#endif
+
 // Turbo Module Name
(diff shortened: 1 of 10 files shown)


TheTrainGoes:
Hi @sirrow ,

What is the use case for this change?

sirrow:
This is useful for many game player who want to get prompt input when pushing turbo enabled button.
Generally this is useful for SHMUP games. In current implementation input for turbo button delays if bTurboFlicker==true. So if shoot is assigned on turbo button, shoot will delay.
My exact case is.. I'm player of Mr.Driller, an action puzzle game. In this game, input delay of turbo button can be cause of time consuming or gameover.

arntsonl:
Closed and merged again with PR#549! Thanks @sirrow

PR #531, opened 2023-09-20
lib/ADS1219: use rp2040 i2c directly

sm00th:
Drop the Bitbang_I2C dependency and use rp2040 i2c directly. This is pretty straightforward as only read/write and init functions are used which are rather small.

Unfortunately this is only build-tested as I don't have ads1219 hardware at hand.

### DIFF
--- a/lib/ADS1219/ADS1219.cpp
+++ b/lib/ADS1219/ADS1219.cpp
@@ -3,18 +3,12 @@
 #include <cstring>
 
 ADS1219::ADS1219(int bWire, int sda, int scl, i2c_inst_t *picoI2C, int32_t speed, uint8_t addr) {
-  bbi2c.iSDA = sda;
-  bbi2c.iSCL = scl;
-  bbi2c.picoI2C = picoI2C;
-  bbi2c.bWire = bWire;
-  iSpeed = speed;
-  address = addr;
+  i2c = new I2C(picoI2C, sda, scl, speed, addr);
   config = 0x00;
   singleShot = true;
 }
 
 void ADS1219::begin() {
-  I2CInit(&bbi2c, iSpeed); // on Linux, SDA = bus number, SCL = device address
   reset();
 }
 
@@ -26,7 +20,7 @@ void ADS1219::begin() {
 //  resets the digital filter and restarts continuous conversions.
 void ADS1219::start(){
   uc[0] = 0x08;
-  I2CWrite(&bbi2c, address, uc, 1);
+  i2c->write(uc, 1);
 }
 
 // 8.5.3.4 POWERDOWN (0000 001x)
@@ -36,7 +30,7 @@ void ADS1219::start(){
 //  START/SYNC command is issued, all analog components return to their previous states.
 void ADS1219::powerDown(){
   uc[0] = 0x02;
-  I2CWrite(&bbi2c, address, uc, 1);
+  i2c->write(uc, 1);
 }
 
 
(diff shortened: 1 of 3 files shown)


arntsonl:
We have a full rewrite of i2c coming, so this code is all going to get ripped out. Sorry we don't have it publicly documented anywhere but i2c will be rewritten.

sm00th:
Who's doing the rewrite? I talked to BigRed on discord a while ago and floated an idea of writing a proper rp2040-oled library and moving to that instead of hacking up OneBitDisplay. He seemed to be on board with it. I have that part on the way https://github.com/sm00th/GP2040-CE/tree/oledlib, just wanted to add a couple more touches (like double buffer so that gp2040 doesn't re-render the whole screen every frame) before posting it.

mikepparks:
The OLED library can use some work, yes.  I'm actually in the middle of rewriting how we handle I2C as a whole, both to remove the "in name only" dependency on BitBangI2C, as well adding some flexibility and management that we didn't have before.

sm00th:
> The OLED library can use some work, yes. I'm actually in the middle of rewriting how we handle I2C as a whole, both to remove the "in name only" dependency on BitBangI2C, as well adding some flexibility and management that we didn't have before.

Could you please explain a bit more or if you have something in the works could you share the code you are working on? I don't think I fully understand.

The way I see it. Right now there are 2 i2c users: ads1219 and i2cdisplay addons, they both work through a wrapper that is BitBangI2C, which is there only for historical reasons because it was easier to port OneBitDisplay this way.

Now what you describe sounds like another wrapper around pico sdk's i2c functionality that will be used in lieu of BitBangI2C. Is there a need for one? Won't that make using any potential third-party libraries that use pico-sdk directly harder? How would I have to adjust https://github.com/sm00th/GP2040-CE/tree/oledlib to work with it?

mikepparks:
Currently, there are three I2C users: the two you mentioned and the Wii addon. Removing the BitBangI2C wrapper is to remove name confusion with the existing project it was based on, as well as to clearly state were not actually bit banging our I2C usage.

We're in the process of rewriting the wrapper to allow for better sharing of the controller blocks, as well as to allow more flexibility on usage. Currently, every addon that leverages I2C is initializing the controller on their own, which introduces issues with restarting the controller, as well as the potential for speed conflicts. At the pico-sdk level, the SDK is only concerned with interfacing with the controller, and isn't concerned about what sits on top. There's no effort to manage communications by multiple devices.

The idea with the rewrite is to give us an abstraction layer that will:
* handle controller initialization globally, including leveraging both blocks (removes configuration complexity for addons and moves initialization earlier in the boot process)
* improve traffic management for multiple devices on the controller by using a common interface and queuing (there are known conflicts with the display clobbering anything it shares the controller with)
* restrict GPIO pin usage only to the ones shared by the controller (at the moment, any pin can be defined in the web config)

Most libraries that leverage pico-sdk directly assume that they're the only device that will use the controller. In our scope, we'd like to take advantage of I2C's ability to share the controller with multiple devices, so some minor adjustments to the addon libraries may be needed to leverage our HAL.

sm00th:
Thanks for the great write-up. This makes much more sense now. Re-init part should be easily addressable with third-party libs. Traffic control is trickier since third-party libs won't be using your wrappers, but as long as they use blocking version of write/read should be solvable.

Do you have any ETA on this? Not rushing you in any way, just trying to figure out timings.

Oh and if you'll need any help on this - let me know.

Issue #532, opened 2023-09-20
4-Way Joystick Mode behaves differently from expectation


sirrow:
### Prerequisites
- [x] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context
* Firmware Version: v0.7.4
* Name of device: Pico (but the issue should be not depend to device.)
* Is this a custom build?: no

### Description of Issue

I found 4-Way Joystick Mode behaves differently from my expectation.
It seems for me that SOCD and 4-Way Joystick Mode seems conflicted and 4-Way Joystick Mode need to care only diagonal direction but also opposite direction.
It would be grateful if some of you consider if my expectation reasonable.
2 examples I feel strange are below.

##### Preconditions
- SOCD is Last Win
- 4-Way Joystick Mode is enabled

##### Scenario
1. push left
2. push right
3. push down
4. release right

##### Expected Output
down is pressed

##### Actual Output
left is pressed because SOCD outputs left at step 4 and 4-Way Joystick Mode get output of SOCD.

##### Preconditions
- SOCD is Off
- 4-Way Joystick Mode is enabled

##### Scenario
1. push left
2. push down
3. push right

##### Expected Output
right is pressed

##### Actual Output
Both left and right are pressed. Because 4-Way Joystick Mode only seems only checking which of directions which consists diagonal is pressed last and SOCD is Off.

sirrow:
Examples are checked in below environment:
- Input Mode in GP2040-CE
  - Xinput
- gamepad tester (reproduced in both web site)
  - https://hardwaretester.com/gamepad
  - https://pelsin.github.io/gamepad-tester/
- Browser
  - Chrome
- OS
  - Windows 11

bsstephan:
@sirrow - if you get a chance can you try my changes in https://github.com/OpenStickCommunity/GP2040-CE/compare/main...bsstephan:GP2040-CE:four-way-improvements ? I have 4-way running before SOCD now, and it makes Example 1 work as expected. I'm wondering if you can try it out and make sure it didn't break some other expectation, but it looks good to me. I do not know if it addresses Example 2 at all.

[...]

PR #533, opened 2023-09-20
Another 4-Way Joystick Mode implementation

sirrow:
This is quick implementation which meet my expectation described in https://github.com/OpenStickCommunity/GP2040-CE/issues/532.

### DIFF
--- a/headers/gamepad/GamepadState.h
+++ b/headers/gamepad/GamepadState.h
@@ -6,6 +6,8 @@
 #pragma once
 
 #include <stdint.h>
+#include <list>
+using namespace std;
 #include "GamepadEnums.h"
 #include "enums.pb.h"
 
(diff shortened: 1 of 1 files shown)


TheTrainGoes:
Hi @sirrow ,

I will check thit out when I have some time

bsstephan:
@sirrow is it the case that the draft only addresses Scenario 2? I don't have a great way to test SOCD=Off at the moment, but it appears that Scenario 1's behavior is the same with your draft.

sirrow:
> @sirrow is it the case that the draft only addresses Scenario 2? I don't have a great way to test SOCD=Off at the moment, but it appears that Scenario 1's behavior is the same with your draft.

As you @bsstephan described at https://github.com/OpenStickCommunity/GP2040-CE/issues/532#issuecomment-1728300593, conbination of  https://github.com/OpenStickCommunity/GP2040-CE/pull/533 and https://github.com/OpenStickCommunity/GP2040-CE/pull/537 works well.

sirrow:
https://github.com/OpenStickCommunity/GP2040-CE/pull/537 is already merged.
I make this ready for review with current implementation.

I'll write another pull request for revise the implementation for ease of understand, rewrite with std::list.

sirrow:
I revised my implementation with std::list.
New implementation is functionally same as previous one.

Old implementation is moved to https://github.com/sirrow/GP2040-CE/tree/4way-alt-implementation-dev-v1

arntsonl (approved):
Looks good, standard list makes sense if we're doing push/pop.

PR #537, opened 2023-09-20
run 4-way filtering before SOCD cleaning

### LINKED ISSUE #532: 4-Way Joystick Mode behaves differently from expectation

bsstephan:
This addresses Example 1 of #532 without any immediately obvious detriment to the 4-way/SOCD interaction, other than it has an order of operations problem (L > D > R = nothing, but R > D > L = L).

However, #533 fixes that issue by implementing filtering better, and it fixes Example 2 of #532.

@TheTrainGoes is going to do more testing, but in my testing, these two PRs together fix #532.

### DIFF
--- a/src/addons/dualdirectional.cpp
+++ b/src/addons/dualdirectional.cpp
@@ -65,6 +65,7 @@ void DualDirectionalInput::debounce()
 
 void DualDirectionalInput::preprocess()
 {
+    const DualDirectionalOptions& options = Storage::getInstance().getAddonOptions().dualDirectionalOptions;
     Gamepad * gamepad = Storage::getInstance().GetGamepad();
 
  	// Need to invert since we're using pullups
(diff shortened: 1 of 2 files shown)


sirrow:
I tested code on my testing branch merging both https://github.com/OpenStickCommunity/GP2040-CE/pull/533 and https://github.com/OpenStickCommunity/GP2040-CE/pull/537.
Everything works fine with https://github.com/OpenStickCommunity/GP2040-CE/pull/533 and https://github.com/OpenStickCommunity/GP2040-CE/pull/537 about 4-Way Joystick.

arntsonl (approved):
Nice, good little fix for SOCD

PR #538, opened 2023-09-20
Fix error display string template in Ps4.tsx. Add missing Web API import.

vgf89:
I wasn't able to get the PS4 Mode addon to work, it just showed the following error at me after clicking Verify and Save:  'ERROR: Could not verify required files: ${e}'

Turns out that was a string literal instead of a template. Replacing the ' with ` in Ps4.tsx made it a template and enabled printing the actual error.

The actual error was that WebAPI wasn't found. I added the missing import and it started working.


Pelsin:
Great fix, thanks for the PR!
The missing webApi import was probably my bad when separating all the addons from the addonConfigPage.

arntsonl (approved):
Good catch!

PR #545, opened 2023-09-23
Config for the RP2040 Advanced Breakout Board USB Passthrough

TheTrainGoes:
This also boots directly into PS4 mode with the fightstick type and USB setup on GPIO 23/24


arntsonl (approved):
Looks good! I'll pull in the other change as well so ps passthrough works correctly.

arntsonl (approved):

PR #549, opened 2023-09-25
Turbo flicker fix from sirrow PR#529

arntsonl:
This is the turbo flicker fix that Sirrow committed without the web config bits.

This is a bug-fix when turbo is released, and then pressed again.


arntsonl (commented):
Confirmed working!

PR #550, opened 2023-09-27
fix 0% dead zone

NickGuyver:
fix for dead zone if user sets it to 0%


arntsonl (approved):
Makes sense, easy fix thank you!

Issue #406, opened 2023-07-11
Gyroscope/Accelerometer support for PS4, PS3, and Switch modes


vgf89:
### Have you checked the GP2040-CE documentation [HERE](https://gp2040-ce.info/#/) to confirm this feature doesn't already exist?

Yes

### Which problem is this feature request solving?

PS4, PS3, and Switch controllers support gyroscope and accelerometer aka IMU inputs. Many popular Switch games use it (Zelda BOTW/TOTK where it's required to complete some puzzles, Splatoon for precision aiming, optional Mario Kart steering control) and a handful of PS3 and PS4 titles do as well. Steam Input in particular allows heavy customization of controllers including gyro aiming, which makes PC mouse-based first and third person shooters enjoyable. GP2040-CE does not have any way of collecting or packaging motion data.

### Describe the solution you would like to see

It looks like analog controls are being actively worked on (which is awesome BTW!), but adding motion controls as well would make it fairly trivial to create fully-featured handheld controllers for Switch, PS3, and PS4 systems.

Adding motion to PS4 mode especially would help in making more featureful controllers for PC/Linux(Steam Deck)/Steam Input since that could support analog triggers and a touchpad as well. It might be easiest to add motion to this mode as it's currently just padded out (or in the mystery field I guess) in the report descriptor: https://github.com/OpenStickCommunity/GP2040-CE/blob/13d8daa41eea633da0f23981d992554def5931e1/headers/gamepad/descriptors/PS4Descriptors.h#L129

Looks like the descriptor for PS3 cuts off before it gets to the motion data. I'm pretty unfamiliar with USB/HID descriptors. How easy would it be to extend this to include motion data? https://github.com/OpenStickCommunity/GP2040-CE/blob/13d8daa41eea633da0f23981d992554def5931e1/headers/gamepad/descriptors/HIDDescriptors.h#L122

I don't know enough about Switch to have any suggestions there and its descriptor looks... more barebones/less documented than the other ones? Or I'm just not looking in the right places. Anyways I've found some relevant discussion as well as a what appears to be a PR for a complete Switch Pro Controller emulator for Arduino Pro Micro. I've linked all of that below.

Of course we would also need support for a 6DoF IMU chip for this any of this to be usable. It appears that many of them support both I2C and SPI. For starters, Switch controllers supposedly to use an LSM6DS, for which Sparkfun has a breakout board, documentation, and a tutorial.

References:
[...]

PR #551, opened 2023-09-28
Add MPU6050 addon for IMU support in PS4 Mode

vgf89:
#406

This only adds IMU support to PS4 mode. ~~For Windows and Linux to recognize the IMU data (in Steam Input, GameConTest, etc), I also had to switch the vendor and product ID to official ones since the panthera one didn't work. A real PS4 doesn't care though, it worked both ways.~~

I don't know enough about how Dualshock 3/SixAxis or Switch controllers work to add support there, but that could be a future issue. Perhaps someone with better HID knowledge would be able to get it to work.

To test this, you'll want an MPU6050 breakout board, also called a GY-521.


dogtopus:
Are you able to test whether the IMU is recognized correctly on a PS4? Seems like the (suspected) IMU parameters in the config report (0x03) are still 0.

I'm not sure how would one test this TBH and this is one of the blockers for me when I make my own DS4 emulator.

vgf89:
I've tested at least the accelerometer on the PS4 and it worked fine for pitch and steering in Wipeout Omega Collection, just like an official controller. I don't have any games on PS4 with gyro aim, though if there are any cheap ones to test with I could do that. However, with PC through Steam Input (and GameConTest), gyro as mouse works perfectly and has the same speeds as my PS4 controller.

I suspect that part of the hid descriptor (or possibly even the entirety of it when using the official vendor id/product id) just gets ignored/overridden when using the official vendor and product id, but the hid descriptor being incomplete could explain the lack of motion while using different IDs. I think I'll play around with that a little bit more, but it works as is.

vgf89:
I did a little more testing on the PS4 side. Looks like we actually need to use the Panthera VID/PID for PS4, but we have to use DS4 ones on PC. PS4 doesn't recognize the controller if we're using the DS4 VID/PID. Maybe some HID magic can be done to fix PC IMU support without breaking PS4 support while on third party IDs?

Anyways, the accelerometer seems slightly less sensitive in a split screen comparison Wipeout Omega Collection (probably just due to lack of calibration, i.e. my up axis reads 8.5G while a real DS4 reads 9.8G) but it's still completely playable. The gyro seems comparable as well in TLoU P2, but since I don't have any split-screen games to compare with it's a little harder to tell. On PC I recall it being very close when doing tests with Steam Input and seeing how far a 90 degree controller turn with high sensitivity settings would spin me in Quake.

To fix any inaccuracies, we'd need a comprehensive calibration sequence. 3 steps for the accelerometer (to determine offset and scale for every axis compared to gravity) and also 3 steps for the gyro (probably rotating the controller 360 around every axis to determine the scale). But at a base, just calibrating the gyro offset once (which I'm already doing) makes everything playable, just not quite flawless.

arntsonl:
I'm going to switch this over to draft mode. I have an MPU6050 breakout board on me to test this, but we can look at how we could add this as an experimental option.

vgf89:
Got third party vidpids to work with motion. The changes to the 0x03 feature report in ps4_driver.cpp are currently hardcoded. I'll probably switch the internal state values back to G and deg/s floats again (instead of raw sensor values) and just choose sensible conversions/scale factors for the ps4 driver, which would be a lot less convoluted than trying to store the scale factors themselves in gamepadstate or something (especially once we add Switch and PS3 compatibility)

arntsonl:
Going to close this draft and have you resubmit when its ready to be reviewed. Please feel free to ping any of us in discord for questions or ideas, looking great so far.

PR #554, opened 2023-10-01
Implement Macro feature

### LINKED ISSUE #235: Bounty - Macros V2.0
This is a bounty for another stab at Macros.  

Previously we had a working PoC for Macros where the pushing of the BootSel button on a raw Pico could be used to input a series of defined button presses in a pre-defined order.

This was tested but we never got further on it.  

I would like to explore Macros v2.0 where a file containing macro information can be uploaded.   

An example of the structure could be as follows:

Input // on or off // pause for next line
DOWN,on,0.05,
RIGHT,on,0.05,
DOWN,off,0.05,
RIGHT,off,0.05,
p1,on,0.05,
P1,off.0.05,

In this example a Hadouken for Street Fighter is performed.  

The macro button when pressed should start the macro.  Pressing the macro button a second time during the macro should stop it.

In the web-config there should be an option to either have the macro trigger once per button press or loop until the macro is stopped by having the macro button pressed again.

deeebug:
As requested, the macro addon has been implemented and is here in this PR.

Important thing to note is that the allocated flash space has been raised from 8K to 16K.
This addon supports 6 macros with 50 input lines each which also allows for waiting before moving on to the next line. This could be updated in the future, as needed.

Room for improvement is plenty. The logic for macro itself could be significantly optimized, and is frankly dumb. Same can be said for both UI and UX. But right now where it is at seems like a good foundation.

Thanks to @TheTrainGoes and @InfraredAces for testing and feedback.
Extensive testing is still pending, so reviews are appreciated.

Closes #235.


InfraredAces:
Testing this [Action run](https://github.com/OpenStickCommunity/GP2040-CE/actions/runs/6375211054) I ran into issue with setting a macro. I was able to set the Macro Pin, enable Macros, and save. However, when I set a specific macro as "Activated", it would not save and threw an error in the browser console. I've included screenshots below.

image


image

deeebug:
Thank you for testing! These should be fixed now.

arntsonl (approved):
Incredible work, thank you @deeebug !! Works great on my setup

PR #555, opened 2023-10-03
Moving report counter back into PS4 with proper 6-bit roll-over

arntsonl:
There's been observable drops in inputs when report counter is missing in the PS4 reporting structure.

Playing "Ultimate Chicken Horse" on PS5, it is observed that rapid inputs will eventually cause the gamepad to lock when report counter is missing.

Still confirming adding report counter back in fixes the issue, but it appears it is important somewhere in the gamepad stack on the PS4/PS5.


arntsonl (commented):
This issue has now been replicated in Final Fantasy 2, Corpse Party, and Ultimate Chicken Horse. Until we can narrow this down further, report counter is coming back in.

PR #562, opened 2023-10-08
Switch Press macro type to positive edge

deeebug:
As discussed, this makes the Press type macro on positive input instead of the previous negative edge.


arntsonl (approved):
Looks great!

PR #565, opened 2023-10-09
Add Macro Lock to Focus Mode addon

deeebug:
This implements a switch to lock the macro feature to the Focus Mode addon.


arntsonl (approved):
Looks great!

PR #570, opened 2023-10-11
Change target button for DirectInput/PS3 Mode in documentation

outphase:
Currently both Xinput and DirectInput/PS3 Modes are labeled in usage.md as being invoked by B2. This is to change the DirectInput/PS3 button to be B3.


arntsonl (approved):
Good catch!

PR #571, opened 2023-10-12
Update analog trigger mode to include digital state

mikepparks:
Allow digital input state and analog value state to co-exist for triggers when Gamepad.hasAnalogTriggers is true.
Added analog trigger check for keyboard host to avoid potential issues.


arntsonl (approved):
Good catch on this one. I would have had no idea we didn't treat Left/Right trigger the same on X-Input.

PR #576, opened 2023-10-15
Fix Reflex CTRL SNES Pinout, Rename Reflex Encode, Add Reflex Encode V2

misteraddons:


TheTrainGoes (changes requested):
Hey @misteraddons ,

Looks mostly good.  Line 13 in the v1 and v2 does not need to specify all of these. 

You can take a look here if want to foce a boot to a specific mode:
https://github.com/OpenStickCommunity/GP2040-CE/blob/0589f43e704d80b4bd81a19fc2dd35b88b13d45e/configs/RP2040AdvancedBreakoutBoardUSBPassthrough/BoardConfig.h#L66C1-L66C42

Otherwise you can leave as is and it will default to Xinput.

misteraddons:
Awesome, I just pushed an update. Thanks for the tip!

TheTrainGoes (approved):
Looks good!

arntsonl (approved):

PR #583, opened 2023-10-18
Updating All Gamepad MID Macros to use our mid-function

arntsonl:
We might think about a better way to do this in Gamepad, like "state->mid()" or something but for now this works great!


arntsonl (commented):
Nothing broke!

arntsonl (commented):
Slightly updated indents to match code from BRP

PR #582, opened 2023-10-18
fix joystick midpoints

NickGuyver:
Create helper function and add fix to missed areas.

PR #577, opened 2023-10-16
Tilt Input add-on LS/RS normalisation

InfraredAces:
This PR changes the following things with the Tilt Input add-on

- Make both Right Analog Stick and Left analog stick behave identically to one another compared to the current implementation
- Changes how the tilt factor values entered into the web configurator alter the magnitude of the direction input
- Adds documentation for the new behavior
- Adds input validation for the Tilt Factor Values, limiting them to values 0-100
- Sets default values in add-on for all values, except the Pico config to 0
- Sets default values of Pico Config to match the original values prior to custom values (T1 - 35%, T2 - 65%)

### Left Analog Stick

- Left Analog Stick (LS Tilt 1 Factor XY = 35) - Input Left while holding Tilt 1 => Send input at 65% Left
- Left Analog Stick (LS Tilt 1 Factor XY = 35) - Input Up while holding Tilt 1 => Send input at 65% Up
- Left Analog Stick (LS Tilt 2 Factor XY = 65) - Input Left while holding Tilt 1 => Send input at 35% Left
- Left Analog Stick (LS Tilt 2 Factor XY = 65) - Input Up while holding Tilt 1 => Send input at 35% Up

Pressing tilt with a Tilt Factor Left = N results in a (100-N) so a 35 Tilt Factor value results in a 100-35%=65%.

### Right Analog Stick

The X-axis of the Right analog stick behaves the same as the Left stick, but the Y-axis behaves in a different manner. At 0 Tilt Factor Right Y, RS would go all the way to the top, and at 200, it will go all the way to the bottom. 100 will give you the middle point.

Up and down are entirely unaffected by either T1 or T2 and pressing T1 or T2 + Left or Right will ignore the Up or Down inputs and treat it as if you only input T1 or T2 + Left or Right.

Current Documentation - `The Right stick shifts down for Tilt 1 and up for Tilt 2 by default.`

This behavior **only** occurs with the specific values that are not set as default anywhere other than the Pico config, otherwise the default values are set to -1. Other values will have behavior depended on the actual values used.

#### "Default" Behavior

- Right Analog Stick (RS Tilt 1 Factor X = 35, Tilt Factor 1 Y = 170) - Input only Left while holding Tilt 1 => Send input at 65% Left, 65% Down
- Right Analog Stick (RS Tilt 1 Factor X = 35, Tilt Factor 1 Y = 170) - Input only Right while holding Tilt 1 => Send input at 65% Right, 65% Up
[...]

PR #585, opened 2023-10-18
Tilt state refactor and analog mid adjustments

mikepparks:
Reworked @InfraredAces #577 to include the recent Analog middle value changes from #582.  Also refactors the state checking when pressing the modifiers to remove redundant logic.


arntsonl (approved):
Looks great!

PR #587, opened 2023-10-19
Fixes Macro type Press and Loop issue

deeebug:
This fixes:
1. Resetting issue introduced with macro type Press when switching to positive edge trigger.
2. First input line only held for a frame upon looping.


arntsonl (approved):
Yeah debounce is a pain

PR #593, opened 2023-10-22
Update to the default config layout

TheTrainGoes:
The update changes the default display layout so that the word 'SOCD' remains while also allowing for the macro M and clear identifiers for DPad / Left Stick / Right Stick


arntsonl (approved):
Looks good!

PR #596, opened 2023-10-24
Updated info for the Reflex Encode boards

TheTrainGoes:


arntsonl (approved):
Rename looks good

arntsonl:
Looks good! This will require a doc update on our next release as we're going to be out of sync with cmake and our 0.7.5 downloads.

Issue #488, opened 2023-09-01
Investigate PS4 Report Counter Latency


arntsonl:
### Prerequisites

Please check the following before posting an issue / bug report.

- [x] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: 0.7.4-dev
* Name of device: Pico
* Link to where you purchased the device:
* Is this a custom build?:
* Operating System:
* Browser:

### Expected Behavior

Adding report counter increment to PS4 data report should NOT increase latency

### Current Behavior

Adding report counter increment to PS4 data report does increase latency????

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Modify report counter in the PS4 report
2. Watch latency go up 1ms??

### Screenshots & Files

This will require some USB capturing and looking at the values. Maybe the ++ was overflowing our 6-bit value, maybe this value is just never used? We really need to pull USB captures for third-party devices and see how they handle the report overflow issue.

jfedor2:
Here's what the main loop looks like:
```
while True:
    button_state = fetch_button_state()
    update_input_report(button_state)
    if tud_hid_ready() and (report != previous_report):
        tud_hid_report(report)
        previous_report = report
```
This is executed multiple times per USB frame.

Once tuh_hid_report() is called, the report that's going to be sent to the host at the next start-of-frame packet is locked in. Anything that happens after that (even before the start-of-frame packet) will only be taken into consideration on the next frame. If you press button A at t+0.2ms and button B at t+0.6ms, an input report with A pressed will be sent at t+1ms and another input report with B pressed will be sent at t+2ms.
![counter1](https://github.com/OpenStickCommunity/GP2040-CE/assets/2271721/fdd6f7a2-de38-4e08-b7e5-107cff46d51e)
As you can see this is already a problem even before we start considering the counter. With the counter the main loop looks like this:
```
while True:
    button_state = fetch_button_state()
    update_input_report(button_state)
    report.counter += 1
    if tud_hid_ready() and (report != previous_report):
[...]

PR #599, opened 2023-10-25
Sychronize send_report with host SOF packet. Should fix main problem mentioned in issue #488

vgf89:
This change makes it so that we don't attempt to run send_report unless the host has already sent an SOF packet. This should prevent prematurely locking in outdated inputs and delaying additional inputs in quick success (sub 1ms between button presses). This should also prevent introducing additional latency with noisy analog inputs or future additions such as motion data.

I unfortunately do not actually have the hardware set up to do latency tests for quick-succession button presses. What I can say is that the polling rate and jitter in XInput mode has not changed at all, as tested with XInputTest, when having the gamepad state change every single frame.

### DIFF
--- a/lib/TinyUSB_Gamepad/src/hid_driver.cpp
+++ b/lib/TinyUSB_Gamepad/src/hid_driver.cpp
@@ -63,4 +63,4 @@ const usbd_class_driver_t hid_driver = {
 	.open = hidd_open,
 	.control_xfer_cb = hid_control_xfer_cb,
 	.xfer_cb = hidd_xfer_cb,
-	.sof = NULL};
+	.sof = sof_callback};
--- a/lib/TinyUSB_Gamepad/src/hid_driver.h
+++ b/lib/TinyUSB_Gamepad/src/hid_driver.h
@@ -10,6 +10,7 @@
 #include "gamepad/descriptors/SwitchDescriptors.h"
 
 extern const usbd_class_driver_t hid_driver;
+extern void sof_callback(uint8_t rhport, uint32_t frame_count);
 
 bool send_hid_report(uint8_t report_id, void *report, uint8_t report_size);
 bool send_keyboard_report(void *report);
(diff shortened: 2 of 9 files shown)


jfedor2:
I'm not sure if it's safe to do it this way as we have no way of knowing how soon the host will poll for the input report after the SOF packet and at that point the data already has to be waiting. Probably better to treat the SOF packet as a reference and measure 900us (or whatever) from that point in time, to be sure all of our processing finishes safely before the data is needed.

vgf89:
Now we keep track of the last SOF time and call send_report in ~900us. This should be safer

arntsonl:
Going to close this one until we get some evidence that adding SOF helps with latency/input issues. USB is a tricky one, TinyUSB tries to get as close as possible but I'm sure there are things we can work on. Would be a really cool experiment to see if you can improve the PIO stuff since that gives us really accurate reads.

arntsonl:
Some good evidence that this could be hitting first frame on PS4!!

jfedor2:
I've given it some more thought and the issue with synchronizing to the SOF packet is that the actual report transfer may not happen immediately after it, but some time later in the frame, depending on what other devices are present on the bus.

We could try synchronizing to the transfer itself, but that might also not be as straightforward. Knowing when the transfer happens shouldn't be a problem, we should be able to use `tud_hid_report_complete_cb`, but the transfer might happen at different times frame-to-frame and not just because devices are plugged or unplugged (that's not a big issue as it happens rarely), but more importantly the exact time might depend on whether or not other devices were transmitting data in this frame (and how much data).

Synchronizing to the transfer time still looks like a viable approach to me, but I think it requires some consideration. Maybe we could synchronize to the earliest time in the frame at which the transfer happened in the last X seconds? That way we'll almost never miss it and we will minimize latency most of the time.

arntsonl:
Going to close this one again as we saw jfedor's results in discord showing its not exactly on the SOF packet.

PR #604, opened 2023-10-29
Rewrite to main and add-ons debouncing

NickGuyver:
This PR addresses a few issues:
- Change the core logic on the game pad debouncer and several add-ons to be more optimized
- Make debouncing code more readable and consistent
- Introduce debounce helper functions for add-ons that could use them

While no change in latency was measurable with my setup, the logic has changed so that presses will always take with no delays. Only release will be delayed, on a per button basis, as currently set at 5ms. Previously, presses and releases were delayed by 5ms based on change equally. Also, checks were added to avoid wasting cycles if the states haven't changed in all the debounced areas.


arntsonl (approved):
Looks good! Gets us part of the way to standardizing our debounce logic

PR #605, opened 2023-10-30
Update with v0.7.5 latency testing

TheTrainGoes:
Tests were done using Ferals optamized testing sketch found here: https://github.com/FeralAI/usblag_optimized/blob/main/usblag_optimized.ino


arntsonl (approved):
Nice! That is great to have both the XLSX and the updated numbers

PR #608, opened 2023-11-03
Add hotkeys for all the buttons.

eulerdisk:
Hi.

I made this pull request for A) completeness, B) it's needed for controllers which have start/select/etc... buttons not very handy.
(I'm building another custom hitbox with a single FN button near to the pinky to hotkey all other secondary buttons to the standard B* buttons. In games like street fighter 6 you need to press a lot of these keys in training mode)

Another possible implementation would be having momentarily profile activation (profile X is active only while pressing button Y, releasing the buttons gives back profile 0, similar to QMK momentarily layers)


bsstephan:
On the surface it sounds a bit silly, but I see the use case and what we have makes this totally possible with almost no effort. The PR is clean and easy to understand. I think this can merge as is.

eulerdisk:
@bsstephan The only question I have is about existing HOTKEY_HOME_BUTTON which apparently is redundant with HOTKEY_A1_BUTTON. The term "home" is used only for Switch and Arcade in the docs. I think it should be removed for cleanliness, but I understand if you want to keep it.

arntsonl (approved):
Looks clean, thank you for the PR!

bsstephan:
> @bsstephan The only question I have is about existing HOTKEY_HOME_BUTTON which apparently is redundant with HOTKEY_A1_BUTTON. The term "home" is used only for Switch and Arcade in the docs. I think it should be removed for cleanliness, but I understand if you want to keep it. The best thing for me is to simply rename HOTKEY_HOME_BUTTON to HOTKEY_A1_BUTTON and remove the new HOTKEY_A1_BUTTON.

Obviously it was good as is and metged, but for the record I agree with the motivation to use A1 instead of Home, but I think some future refactoring may address that concern in a clear fashion. Thanks for the PR!

PR #610, opened 2023-11-05
Fixes to gpioMappings migration and webconfig

bsstephan:
This addresses the following identified issues:
* gpioMappings losing mappings on GPIO 0 and 1 due to erroneous detection of enabled addons
* being unable to un-ASSIGNED_TO_ADDON the PLED pins when setting them to -1 in webconfig
* (re)initializing the deprecated buttonMappings for no reason

It does not address:
* the fact PLEDs are migrated as ASSIGNED_TO_ADDON when not in use because some BoardConfig.h files nevertheless enabled them
* the large growth of fsdata.c


arntsonl (approved):
Good little fixes

PR #613, opened 2023-11-06
Add CNAME to static files

InfraredAces:
Add CNAME to static folder for copying to root of deployment branch


Fortinbra (approved):
Let's go!

PR #614, opened 2023-11-06
Only import SHA256 to decrease fsdata by ~50kb

arntsonl:
Only import SHA256 from crypto-js. (thanks for the tip @Pelsin !!)  Should reduce our webconfig by quite a bit


SavageCore:
Was there an issue with `import sha256 from 'crypto-js/sha256';`? That would be much better if it worked compared to maintaining our own "fork" of the code within webconfig. Anyway, a bit late now, probably!

PR #616, opened 2023-11-06
Recompiled makefsdata with a max. file size of 2 MB

mthiesen:
I have recompiled the Linux version of the `makefsdata` tool to support input files of up to 2 MB in size. The size of the webconfig JS code is dangerously close to the current 1 MB limit.


bsstephan:
We shaved off enough KB that we don't need it right away, but this works for me, so I think we should merge it to future-proof ourselves. Thanks for all the help!

arntsonl:
Yeah this is a good future-proof change

arntsonl (approved):

PR #620, opened 2023-11-11
docToPin: only unset the old pin to NONE if it's a valid pin

bsstephan:
This resolves a webconfig crash where setting an addon pin from -1 to -1 would crash the board until it was powercycled.


mikepparks (approved):
Confirmed fixes the issue that was causing the Display page in webconfig to stop processing events and hang the device after setPreviewDisplayOptions route was called.

arntsonl:
Nice, good fix!

PR #627, opened 2023-11-16
Added default hotkeys to descriptions

outphase:
The home button and SOCD hotkeys did not have their default bindings. They have been added for consistency.


InfraredAces (approved):

InfraredAces:
Merged this in because, even if we remove all of the default hotkeys later, some of them are already in there and this makes it consistent

Issue #602, opened 2023-10-28
Bounty - Re-implement input history


TheTrainGoes:
Bounty - Re-implement input history

Long ago we had a PR for input history that also changed the default layouts on a number of screens (https://github.com/OpenStickCommunity/GP2040-CE/pull/53/files).

I have tested this on our current setup and it works just fine.

Completion looks like:
- Added as a new addon
- Ability to toggle on / off
- Ability to set length of input history via an entry box

Setting the length of the input history will allow people to adjust it based on their displays.

TheTrainGoes:
This work has been completed.

PR #630, opened 2023-11-18
Re-implement input history #602

kiralydavid:
* Added input history as a new add-on.
* You can toggle it on/off from the web-config.
* It's possible to set the length of the history string (in characters) and position it on the screen by setting the starting column and row.


TheTrainGoes:
This looks great!  Thank you @kiralydavid 

I will test this out when I have some time.   Are you on the Discord?

InfraredAces (changes requested):
Documentation piece looks good. 

Thank you for following the template. Though, if you could include the `### Installation` header and just put "None." beneath so as to maintain consistency with the template and other documents.

mikepparks:
A few minor adjustments I'd make, just for the sake of following the existing code patterns.

* I understand the reasoning for doing it, but try to avoid changing the pattern on addon constructors.  Creating the Display addon should not require the Input History addon, even if it's disabled.  
* Instead, I would suggest adding a method on the addon so we only have to attach the two addons if they're enabled.  It should ultimately be optional.
* Check the state of both Input History and Display in your `available()` method if the Display is not enabled.  See the Wii Extension addon ``available()`` for a means of doing this.

site/docs/add-ons/input-history.mdx
InfraredAces, line 1: I noticed in testing this add-on that some of the symbols don't exactly match up with the Input tables used for GP2040-CE. Please include a markdown table with all of the characters used to represent the various inputs for each of the input modes (e.g. XInput, Switch, PS4, etc.) under Miscellaneous Notes.

InfraredAces (dismissed):

mikepparks:
Code changes are looking good on my end. Nice turnaround!

arntsonl:
@kiralydavid Looks great! We just had a push to our i2c peripherals, could you pull latest and resolve conflicts on your end? Should be pretty straight forward, @mikepparks did a great job on the i2c stuff.

arntsonl (approved):

PR #634, opened 2023-11-20
JavaScript implementation of makefsdata

mthiesen:
I have reimplemented the `makefsdata` tool in pure JavaScript. This frees us from having to maintain the binary `makefsdata` tool for different platforms, which was always a hassle. Also, the project can now be built on platforms for which we did not previously provide a `makefsdata` binary.

For compression, I used the [pako](https://www.npmjs.com/package/pako) library, which is a port of the [zlib](https://www.zlib.net/) library to JavaScript. The compression is reasonably fast, and it even compresses a bit better than the [miniz](https://github.com/richgel999/miniz) library we used with the C implementation.

This version of `makefsdata` does not have all the features of the original, but it covers everything needed for this project.


www/makefsdata.js
```diff
@@ -1,41 +1,265 @@
...
+	}
+	return results;
+}
+
+function getLowerCaseFileExtension(file) {
+	const ext = file.split('.').pop();
+	return ext ? ext.toLowerCase() : '';
+}
+
+function fixFilenameForC(filename) {
```
Pelsin, line 81: Unsure how this works, super interesting.
Does this return the character if it's a latin character otherwise "_" ?

www/makefsdata.js
```diff
@@ -1,41 +1,265 @@
...
+	fsdata += '#ifndef FSDATA_ALIGN_PRE\n';
+	fsdata += '#define FSDATA_ALIGN_PRE\n';
+	fsdata += '#endif\n';
+	fsdata += '#ifndef FSDATA_ALIGN_POST\n';
+	fsdata += '#define FSDATA_ALIGN_POST\n';
+	fsdata += '#endif\n';
+	fsdata += '#if FSDATA_FILE_ALIGNMENT==2\n';
+	fsdata += '#include "fsdata_alignment.h"\n';
+	fsdata += '#endif\n';
+	fsdata += '\n';
```
Pelsin, line 159: Wanted to mention template literals or tagged templates as alternatives to string concatenation.
They allow multi-line strings and string interpolation.

Example of template literal
```javascript
let fsdata = `#include \"fsdata.h\"

#define file_NULL (struct fsdata_file *) NULL

#ifndef FS_FILE_FLAGS_HEADER_INCLUDED
#define FS_FILE_FLAGS_HEADER_INCLUDED 1
#endif
#ifndef FS_FILE_FLAGS_HEADER_PERSISTENT
#define FS_FILE_FLAGS_HEADER_PERSISTENT 0
#endif
/* FSDATA_FILE_ALIGNMENT: 0=off, 1=by variable, 2=by include */
#ifndef FSDATA_FILE_ALIGNMENT
#define FSDATA_FILE_ALIGNMENT 0
#endif
#ifndef FSDATA_ALIGN_PRE
#define FSDATA_ALIGN_PRE
#endif
#ifndef FSDATA_ALIGN_POST
#define FSDATA_ALIGN_POST
#endif
#if FSDATA_FILE_ALIGNMENT==2
#include \"fsdata_alignment.h\"
#endif

`
```

www/makefsdata.js
mthiesen: Basically yes, it converts all characters that are not valid for C identifiers to "_". The original `makefsdata` has a function of the same name that does this as well.

I'll add a clarifying comment.

www/makefsdata.js
mthiesen: I considered this, but ultimately decided against it because I wanted to have explicit control over the type of line-endings used. I don't want the line-endings of the source file to "pollute" the generated file. Depending on someone's Git configuration, they might convert line endings to CRLF on checkout, which would then generate `fsdata.c` files with mixed line endings.

Pelsin: Oh that makes sense, clever!

www/makefsdata.js
Pelsin: Thanks for explaining, tried it a bunch to get an understanding.
Should have asked in discord though.

A nit is that :
```
else if (c === '_') {
			varName += c;
		}
```
Can be removed as the default would cover it.

www/makefsdata.js
```diff
@@ -1,41 +1,265 @@
...
+	} else if (Buffer.isBuffer(value) || value instanceof Uint8Array) {
+		value.forEach((byte) => {
+			hexString += '0x' + byte.toString(16).padStart(2, '0') + ',';
+			if (++column >= hexBytesPerLine) {
+				hexString += '\n';
+				column = 0;
+			}
+		});
+	} else {
+		assert(false, 'Invalid value type');
```
Pelsin, line 128: A recommendation would be to not use assert, it's usually for test assertion/test suits.
It would throw some error that might be confusing like: 
```
 actual: false,
 expected: true,
```

instead create an error and throw it:
``` javascript
throw new Error('Invalid value type')
```

Pelsin:
Tested, seems to work nicely!

arntsonl (approved):
I'll wait for resolve-conditions on @Pelsin 's requests before I close this, but looks fantastic. Thank you @mthiesen for doing this, should make it much easier to have everything in Javascript!

www/makefsdata.js
mthiesen: Changed with the latest commeit.

arntsonl (approved):
Looks good!!

PR #635, opened 2023-11-22
User-definable input modes on boot

mikepparks:
This implementation expands on the current boot mode functionality and allows the input types to be user defined. It assumes the current defaults, but adds a new "Boot Input Modes" section in the Settings panel to define up to 8 input modes on main action buttons (B1-4 and L/R 1+2).


arntsonl (approved):
Looks great!! This will make boot options so much better as we add more input modes :D

PR #637, opened 2023-11-25
Input font glyphs

mikepparks:
Added font glyphs for the following inputs:

* General - UDLR arrows
* PS - Triangle, Circle, Cross, Square, Touchpad, PS button, Share
* Xinput - View, Menu, Xbox button
* Switch - Capture, Home


andrepapoti (commented):
These look like they are meant to be used in the input history feature, could you add diagonal directions as well for the arrows so that an horizontal direction + a vertical one can be concatenated in a single glyph ?

arntsonl (approved):
Nice, that's great we can move to fonts.h and have the directional glyphs!

PR #638, opened 2023-11-25
Updated "not available" peripheral messages to include links to config

mikepparks:
Also includes an update to USB context message to remove placeholder description.


arntsonl (approved):
Yeah that's a clean way to do it.

PR #641, opened 2023-11-27
update

TheTrainGoes:
This updates the latency statement to reflect the PS4/5 numbers


arntsonl (approved):
Easy change and makes it honest on our PS4/PS5 latency reporting.

PR #645, opened 2023-11-28
Haute42 support

TheTrainGoes:
Support for the current 6 Haute42 devices which all run off the same config

Haute42 Mini (DIY + assembled)
Haute42 G12
Haute42 G13
Haute42 G16
Haute42 T13
Haute42 T16


configs/Haute42/CMakeLists.txt
arntsonl, line 1: We don't need this file

arntsonl (changes requested):
One file to delete then its in!

arntsonl (approved):
Looks good!

PR #651, opened 2023-12-05
refactor profile switching to be managed by the GP2040 object

bsstephan:
GPIO fiddling was in the Gamepad object, which was fine, but it's going to make more sense in the future for the GP2040 object to manage it, since it has references to addons which will be affected by profile changes, pending another refactor.

this has us essentially initialize/deinitialize "normal" GPIO pins in the higher object and then gamepad and addons just need to worry about button masks or whatever, regardless of the pin initialization itself

### DIFF
--- a/src/storagemanager.cpp
+++ b/src/storagemanager.cpp
@@ -47,7 +47,7 @@ Storage::Storage()
 	critical_section_init(&animationOptionsCs);
 	ConfigUtils::load(config);
 
-	setFunctionalPinMappings(config.gamepadOptions.profileNumber);
+	setFunctionalPinMappings();
 }
 
 bool Storage::save()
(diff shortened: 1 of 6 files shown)


src/storagemanager.cpp
```diff
@@ -138,19 +138,17 @@ void Storage::ResetSettings()
...
 }
 
-void Storage::setFunctionalPinMappings(const uint32_t profileNum)
+void Storage::setFunctionalPinMappings()
 {
 	for (Pin_t pin = 0; pin < (Pin_t)NUM_BANK0_GPIOS; pin++) {
 		functionalPinMappings[pin] = this->config.gpioMappings.pins[pin].action;
 	}
-	if (profileNum < 2 || profileNum > 4) return;
+	if (config.gamepadOptions.profileNumber < 2 || config.gamepadOptions.profileNumber > 4) return;
```
arntsonl, line 149: Can this ever happen? or should it be if (config.gamepadOptions.profileNumber ==1) return; since we have checks earlier?

arntsonl (changes requested):
One code clarity question

src/storagemanager.cpp
bsstephan: Just a sanity check since we're doing array stuff. Shouldn't happen but I guess if anyone did a weird POST to the API or edited the protobuf directly it could be a bad value at startup.

arntsonl: Ahhh got it!

arntsonl (approved):

PR #652, opened 2023-12-06
Migrate profiles to use GpioMappings, allowing extra buttons, DDI, and sliders in profiles

bsstephan:
This refactors the profile code to be extra copies of `GpioMappings`, which moved the core code (and now the profiles, with this change) from button-to-pin to pin-to-action mapping. The core change allowed for arbitrary extra buttons and also moved the DDI and slider pin assignments into the core pin configuration, and now the same applies to profiles.

@Pelsin contributed all the UI code, and I am as always thankful for that. Note that the profiles have moved to the pin mapping page, and now the pin mapping page has four tabs, for the core config + 3 extra profiles.

CAVEATS:
* The ability to remap buttons extends to the Function button! If users remap away their Function button, and are using it for profiles, they won't be able to hotkey their way back to another profile and will have to go into the webconfig to switch profiles again. This seems like a corner case/fair trade to me, so the code takes no particular precaution here.
* The same applies to the buttons critical to boot hotkeys! If you map away your S2, you won't be able to get into webconfig unless you nuke the config or use something like the forced webconfig boot .uf2.
* DDI and the slider add-ons must still be enabled for their pin mappings to matter; the caveat here is that the add-ons are enabled globally, despite profiles, so if you have the slider enabled, its default option (or active pins) override the hotkey-changeable controller setting. If you have a profile enabled that *doesn't* have slider pins, your only option in that profile is the default option. This also seems like a corner case to me, but maybe a future refactor can make this easier on people.

Getting this one out there for people to play with and review.

### DIFF
--- a/www/src/Locales/en/PinMapping.jsx
+++ b/www/src/Locales/en/PinMapping.jsx
@@ -6,7 +6,7 @@ export default {
 		"Mapping buttons to pins that aren't connected or available can leave the device in non-functional state. To clear the invalid configuration go to the <1>Reset Settings</1> page.",
 	'pin-header-label': 'Pin',
 	'profile-pins-warning':
-		'Try to avoid changing the buttons/directions used for your switch profile hotkeys, or else it will get hard to understand what profile you are selecting!',
+		'Try to avoid changing the buttons and/or directions used for the switch profile hotkeys. Otherwise, it will be difficult to understand what profile is being selected!',
 	errors: {
 		conflict: 'Pin {{pin}} is already assigned to {{conflictedMappings}}',
 		required: '{{button}} is required',


Pelsin:
Nice! As a tiny addition, it's also possible to map all profiles with the gamepad record thingy.

Here is a preview on the ui with the base + profiles merged:
Skärmavbild 2023-12-06 kl  16 39 38

www/src/Locales/en/PinMapping.jsx
```diff
@@ -3,8 +3,10 @@ export default {
 	'sub-header-text':
 		'Use the form below to reconfigure your button-to-pin mapping.',
 	'alert-text':
-		"Mapping buttons to pins that aren't connected or available can leave the device in non-functional state. To clear the the invalid configuration go to the <1>Reset Settings</1> page.",
+		"Mapping buttons to pins that aren't connected or available can leave the device in non-functional state. To clear the invalid configuration go to the <1>Reset Settings</1> page.",
 	'pin-header-label': 'Pin',
+	'profile-pins-warning':
+		'Try to avoid changing the buttons/directions used for your switch profile hotkeys, or else it will get hard to understand what profile you are selecting!',
```
arntsonl, line 9: Grammar change:

Try to avoid changing the buttons and/or directions used for the switch profile hotkeys. Otherwise, it will be difficult to understand what profile is being selected!

Pelsin: Awesome, thanks! Will change!

www/src/Components/Navigation.jsx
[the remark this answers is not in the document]
Pelsin: This is a leftover from an earlier version of react-router.

It made the route match exactly if you had multiple routes / descendants starting with the same name.
In v6+ this is the default behavior, so it should be redundant.
More info: https://github.com/remix-run/react-router/issues/4958#issuecomment-582030578

Currently the param/argument does not do anything except gives type errors, i'm trying to fix type errors when doing other changes in a file so in the future we can run type validation before build process.

Skärmavbild 2023-12-06 kl  22 47 25

PR #653, opened 2023-12-07
Enables linting for typescript, a few bug fixes

Pelsin:
Tons of fixes for type/lint errors
I will do a separate PR with fixes that could affect functionality , this PR should be mostly formating and removing of unused code.

### A few bugs became apparent when going through the lint errors
Fixed:
- Chinese flag did not render correctly
- The latest version of the firmware is not shown and the button to download it is also never shown.

Not fixed in this PR:
- Exporting/importing does not work

Removes around 120 errors, but still a few left. One step at a time :)
Skärmavbild 2023-12-07 kl  18 48 54


arntsonl (approved):
Good fixes! That's a lot of catches, nice.

PR #654, opened 2023-12-07
Docs update for merging profile with pin mapping

Pelsin:
Update .gitignore so that we now version control package-lock.json, this should make more deterministic deploy behavior.


arntsonl (approved):
Looks great, .gitignore is much cleaner!

PR #656, opened 2023-12-08
export/import now functioning again, also handles pins and profiles options

Pelsin:
Added a notice about pin exports prior to v0.7.6 will not be saved. 
Skärmavbild 2023-12-08 kl  19 01 27

An improvement to this would be to migrate to the new structure if an older files gets imported.


arntsonl (approved):
Looks good!

PR #658, opened 2023-12-08
Fix issue 487

NickGuyver:
Clean up some old code to get the hotkeys that were listed in issue 487 working correctly. Home was already working when I tested.


arntsonl (approved):
Good cleanup, looks good!

PR #660, opened 2023-12-08
Reconstructed menu organization to prioritize navbar on mobile

marsherino:
These commits restructure the website sidebars in order to make the navbar the default sidebar on the mobile version of the site.

The "Home" sidebar is renamed to "Getting Started" and is disassociated from the homepage. On the desktop site, no sidebar will appear on the homepage until one of the navbar options is clicked. All other manu organization remains the same.

This PR also makes a small adjustment that causes menus to auto-collapse when a new menu is opened, though all menus that are presently opened by default will remain so until one is closed.


arntsonl:
We're going to close this one as this changes our site too much. We should definitely sketch up some pillar pages without the sidebar @Pelsin and @InfraredAces but hiding it causes misalignment and things don't look right.

PR #661, opened 2023-12-08
add more hotkeys

NickGuyver:
Add more hotkeys as requested in issue 503.


arntsonl (approved):
Looks good to me! going to have @bsstephan give it a once over since he did the hotkeys originally

bsstephan:
I don't have a chance to try it out right now, but looks like what I'd expect. I was thinking that it might be good to refactor hotkeys into another repeated protobuf/array thing, but that can happen whenever. This looks good to me.

PR #663, opened 2023-12-13
Use git describe to generate binary version dynamically

bsstephan:
This makes some changes to the build system to replace the hardcoded e.g. "v0.7.5" with the output of `git describe --tags --always`, meaning that the version number bump is more, if not completely, driven by the tagging of e.g. v0.7.6.

`git describe` also, for non-tag builds, adds a count of revisions from a tag, and the commit git SHA, to the string, which is pretty helpful for developers like me who have a gazillion builds laying around. Here's an example:

Screenshot_20231213_113924

```
bss@garnet ~/proj/GP2040-CE % picotool info build/GP2040-CE_0.7.5.56_RP2040AdvancedBreakoutBoard.bin
File build/GP2040-CE_0.7.5.56_RP2040AdvancedBreakoutBoard.bin:

Program Information
 name:      GP2040-CE
 version:   v0.7.5-56-g8f683ae
 features:  double reset -> BOOTSEL
```
(Note both the slightly more descriptive version number in the filename and the descriptive pico version number.)


Pelsin:
A heads up that this might result in the UI always showing there is a new version to download as we will compare "v0.7.5-xxxx" with "v0.7.5"

https://github.com/OpenStickCommunity/GP2040-CE/blob/main/www/src/Pages/HomePage.jsx#L88

bsstephan:
> A heads up that this might result in the UI always showing there is a new version to download as we will compare "v0.7.5-xxxx" with "v0.7.5"
> 
> https://github.com/OpenStickCommunity/GP2040-CE/blob/main/www/src/Pages/HomePage.jsx#L88

Interesting, maybe there's a quick substring compare I can do, or suppress the check entirely if the version has a non-tag format.

Pelsin:
> > A heads up that this might result in the UI always showing there is a new version to download as we will compare "v0.7.5-xxxx" with "v0.7.5"
> > https://github.com/OpenStickCommunity/GP2040-CE/blob/main/www/src/Pages/HomePage.jsx#L88
> 
> Interesting, maybe there's a quick substring compare I can do, or suppress the check entirely if the version has a non-tag format.

Absolutely

```javascript

currentVersion.split("-")[0]
```

bsstephan:
Thanks @Pelsin for the ideas. Here's the new behavior, which suppresses the download prompt whenever the version has the dev build bits:
image

arntsonl (approved):
Looks good!

PR #665, opened 2023-12-15
Add gif support

### LINKED ISSUE #407: Adding animated GIF support to I2C display splash mode

Yes

I am very frustrated that we can't be able to upload and display animated GIF on I2C in splash mode.

Implement animated GIF support for I2C display splash mode.

I was thinking about create a branch with https://github.com/bitbank2/AnimatedGIF/blob/master/src/AnimatedGIF.h implemented and compile my own GP2040-CE version.

Unfortunately no, I don't have enough free time to work on this.

Maybe in the future, but it will take a long time.

andrepapoti:
## Description

Add support for GIFs as an addon.
The data for it is stored between the emulated EEPROM and the actual code with some margin for both to grow.

## Related issues
- Closes #407

### DIFF
--- a/headers/addons/gif.h
+++ b/headers/addons/gif.h
@@ -1,0 +1,34 @@
+#ifndef GIF_H_
+#define GIF_H_
+
+#define GIF_ENABLED 0
+#define GIF_STARTING_GIF_ADDRESS 0
+#define GIF_NUMBER_OF_FRAMES 0
+
+#define GIF_EEPROM_DISTANCE 0xA000 // Leave room of 40kb for the EEPROM to grow
+#define GIF_CODE_DISTANCE 0x20000 // Leave room of 128kb for the code to grow 
+
+#include "gpaddon.h"
+#include "config.pb.h"
+
+#define GifName "Gif"
+
+class GifAddon: public GPAddon
+{
+    public:
+        virtual bool available();
+        virtual void setup();
+        virtual void preprocess() {};
+        virtual void process() {};
+        virtual std::string name() { return GifName; };
+        static bool setupForWrite(size_t numberOfBytes);
+        static bool writePage(std::string& base64, size_t pageIndex);
+        virtual void checkBeforeDraw(int32_t currentMsTime);
+        SplashFrame * currentFrame = nullptr;
+    private:
+        void setCurrentFrameFor(uint32_t index);
+        uint currentFrameIndex = 0;
+        int32_t lastDrawTimestamp = 0;
+};
+
+#endif
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -195,6 +195,12 @@ message ProfileOptions
 	repeated GpioMappings gpioMappingsSets = 2 [(nanopb).max_count = 3];
 }
 
+message SplashFrame
+{
+	optional int32 delay = 1;
+	optional bytes frame = 2 [(nanopb).max_size = 1024];
+}
+
 message DisplayOptions
 {
 	optional bool enabled = 1;
@@ -681,6 +687,13 @@ message InputHistoryOptions
 	optional uint32 row = 4;
 }
 
+message GifOptions
+{
+	optional bool enabled = 1;
+	optional uint32 startingGifAddress = 2;
+	optional uint32 numberOfFrames = 3;
+}
+
 message AddonOptions
 {
 	optional BootselButtonOptions bootselButtonOptions = 1;
@@ -704,6 +717,7 @@ message AddonOptions
 	optional PSPassthroughOptions psPassthroughOptions = 19;
 	optional MacroOptions macroOptions = 20;
 	optional InputHistoryOptions inputHistoryOptions = 21;
+	optional GifOptions gifOptions = 22;
 }
 
 message MigrationHistory
--- a/src/addons/gif.cpp
+++ b/src/addons/gif.cpp
@@ -1,0 +1,117 @@
+#include "configs/base64.h"
+#include "addons/gif.h"
+#include <hardware/flash.h>
+#include "FlashPROM.h"
+#include "FlashPROM.cpp"
+#include "storagemanager.h"
+
+extern char __flash_binary_end;
+
+bool GifAddon::available() {
+    const DisplayOptions& displayOptions = Storage::getInstance().getDisplayOptions();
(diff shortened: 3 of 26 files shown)


arntsonl:
We are going to wait to evaluate this until after the release just a heads up.

headers/addons/gif.h
```diff
@@ -0,0 +1,34 @@
+#ifndef GIF_H_
+#define GIF_H_
+
+#define GIF_ENABLED 0
+#define GIF_STARTING_GIF_ADDRESS 0
+#define GIF_NUMBER_OF_FRAMES 0
+
+#define GIF_EEPROM_DISTANCE 0xA000 // Leave room of 40kb for the EEPROM to grow
```
arntsonl, line 8: we should not be manually modifying the EEPROM inside of an add-on. If we need to move around flash memory, this should either occur inside of the PROM library we're using, or in a main area. This can also potentially break protobuf.

proto/config.proto
```diff
@@ -195,6 +195,12 @@ message ProfileOptions
 	repeated GpioMappings gpioMappingsSets = 2 [(nanopb).max_count = 3];
 }
 
+message SplashFrame
+{
+	optional int32 delay = 1;
+	optional bytes frame = 2 [(nanopb).max_size = 1024];
```
arntsonl, line 201: all frames together are 1024?

proto/config.proto
```diff
@@ -681,6 +687,13 @@ message InputHistoryOptions
 	optional uint32 row = 4;
 }
 
+message GifOptions
+{
+	optional bool enabled = 1;
+	optional uint32 startingGifAddress = 2;
```
arntsonl, line 693: yeah again modifying the address of EEPROM inside of an add-on will definitely break things.

src/addons/gif.cpp
```diff
@@ -0,0 +1,117 @@
...
+    gifOptions.startingGifAddress = startingAddress + XIP_BASE;
+    gifOptions.numberOfFrames = numberOfFrames;
+    Storage::getInstance().save();
+
+    return true;
+}
+
+bool GifAddon::writePage(std::string& base64, size_t pageIndex)
+{
+    if (get_core_num() != 0)
```
arntsonl, line 72: will this get written by core1 or core0?

src/addons/gif.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    const size_t necessaryMemory = numberOfSectors * FLASH_SECTOR_SIZE;
+
+    if (necessaryMemory > availableMemory) return false;
+
+    const size_t startingAddress = alignedFramesMaxAddress - necessaryMemory;
+
+    while (is_spin_locked(flashLock));
+
+    multicore_lockout_start_blocking();
```
arntsonl, line 54: is this thread safe for when USB is processing? or will this only happen during the web config setup

www/package.json
```diff
@@ -8,6 +8,7 @@
 		"axios": "^1.3.5",
 		"bootstrap": "^5.3.0-alpha3",
 		"formik": "^2.2.9",
+		"gifuct-js": "^2.1.2",
```
arntsonl, line 11: I really wish this wasn't called "gifuct" 👎

src/addons/gif.cpp
```diff
@@ -0,0 +1,117 @@
...
+};
+
+bool GifAddon::setupForWrite(size_t numberOfBytes)
+{
+    if (get_core_num() != 0)
+        return false;
+
+    const uint32_t numberOfFrames = numberOfBytes / sizeof(SplashFrame);
+
+    const size_t framesMaxAddress = EEPROM_ADDRESS_START - XIP_BASE - GIF_EEPROM_DISTANCE;
```
arntsonl, line 27: yeah doing eeprom address modification inside of an add-on is really hacky.

arntsonl (changes requested):
We need a better way to store the GIFs, as well as smoother animations in the SSD logic library.

arntsonl:
I'm going to close this PR as there are too many issues with modifying the EEPROM addresses inside of an add-on.

We should have a new display driver coming out shortly, and we can discuss in firmware-development how to go about adding memory. A lot of boards have 16mb so you could utilize the extra space if the chip is big enough. That's how the other PoC did videos on their gp2040 build.

We can circle back on this one after the display driver comes out but the EEPROM address modifications are going to be a no-go from inside of an add-on.

headers/addons/gif.h
andrepapoti: It's not modifying the simulated EEPROM it's using the unused memory between the EEPROM and where the code sits. One is at the start of the flash memory and the other is at the end. We are writing in middle using as high of as an address as possible while maintaining a safe distance from the EEPROM in case it increases in size, currently it's only using 16kb and I left 40kb more in case it needs to increase in size

proto/config.proto
andrepapoti: No, each frame is 1036 bytes, these 1024 bytes  are the pixels for each one of them

proto/config.proto
andrepapoti: The address of the EEPROM is not being modified with this version of the code this is the flash memory layout

Code | GIF_CODE_DISTANCE| Empty space | Gif stuff | GIF_EEPROM_DISTANCE | EEPROM |

src/addons/gif.cpp
andrepapoti: Core 0 just like any ohter wirte operation on the web config api

src/addons/gif.cpp
andrepapoti: I'm not keen to the details but the functions are the same as the ones that are being used on the EEPROM simulation library used by the project

src/addons/gif.cpp
andrepapoti: No address is being modified

andrepapoti:
>  the EEPROM address modifications are going to be a no-go from inside of an add-on.

The project is using a library that simulates an EEPROM at the end of the flash memory addresses but I'm not writing the gif's into it nor modifying any of it's adresses the only thing that I'm doing with it is reading the constant that stores it's starting address to keep gif data from overwriting those addresses. As I said in a previous comment, this is the proposed flash memory layout in this PR:


```
Code | GIF_CODE_DISTANCE| Empty space | Gif stuff | GIF_EEPROM_DISTANCE | EEPROM
```

PR #671, opened 2023-12-18
Xbox One Support (Passthrough Dongle Required)

arntsonl:
Support for the Xbox One (w/ an attached supported dongle).

This adds support to the Xbox One console as well as Windows in Xbox One mode.

GP2040-CE handles the authentication layer as an entirely man-in-the-middle stack and uses a shared authentication buffer to pass up to 1024 bytes back and forth between the console and dongle.

This commit includes an entirely custom, from-scratch XGIP protocol parser.

Currently support dongles: Magicboots Xbox One, Magic-X

Not supported: real controllers, other dongles. We will try to get these working but this is the initial release.


arntsonl (commented):
Looked everything over, looking awesome!

PR #672, opened 2023-12-18
Config migration and input history additions

mikepparks:


arntsonl (approved):
Works great and fixes our peripherals-save options.

PR #674, opened 2023-12-18
Improve ps4 addon message handling

Pelsin:
- The save message should not randomly disappear anymore
- Error/save message now resets after a new file is uploaded
- Should now only be possible to select one file per input
- Removed tons of lint errors
- The code is now formatted the same as the rest of the code base, hence the big change.

There are more things to improve, but this moves us a step forward.

- [ ] Please someone else also verify the upload still works 🙏


arntsonl (approved):
Great clean-ups thank you, easy win for PS4!

PR #680, opened 2023-12-20
Enable Xbox One auth on any board that has PS auth turned on

arntsonl:
Both PS passthrough and Xbox auth passthrough can be turned on and function just fine


arntsonl (commented):
Adding in board config auth modes for Xbox One and enabled passthrough

PR #681, opened 2023-12-20
Added attributions for original Xbox and mini/classic controller modes

mikepparks:


arntsonl (approved):
Yes! So many folks to thank

PR #682, opened 2023-12-20
Added axis-timing counter (increment 1 every input) to PS4 descriptor

arntsonl:
PS4/PS5 have an axis-timing counter that is 16-bit. We just ignored it but its good to have!


arntsonl (commented):
Confirmed working on a PS5

Issue #668, opened 2023-12-16
Drop down lists in Pin Mappings break the page


jack2game:
### Prerequisites

Please check the following before posting an issue / bug report.

- [x] I am running the latest version of the firmware - built from latest main branch
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: latest main
* Name of device: pico
* Link to where you purchased the device: x
* Is this a custom build?: No
* Operating System: Windows
* Browser: Chrome

### Expected Behavior

Drop down list to display properly in Pin Mapping

### Current Behavior

Webpage would be broken when using drop-down list for PIN09-PIN14 and PIN14-PIN29.

### Steps to Reproduce

Go to Pin Mapping page and expand drop down list for any of the Pins mentioned above.

### Screenshots & Files
(gif, hit play button)
![chrome_NyIXRcIK3t](https://github.com/OpenStickCommunity/GP2040-CE/assets/16070158/973a8bf8-04d4-4f18-be5f-96fb0dbc22fc)

PR #685, opened 2023-12-20
Fix pinMapping dropdown layout issues

Pelsin:
My CSS FU is not strong enough to solve the layout issues with css grid breaking to next column.

Modified it to use bootstrap rows.

This should resolve: #668

### DIFF
--- a/www/src/Pages/PinMapping.tsx
+++ b/www/src/Pages/PinMapping.tsx
@@ -1,17 +1,22 @@
-import React, { useCallback, useContext, useEffect, useState } from 'react';
+import React, {
+	useCallback,
+	useContext,
+	useEffect,
+	useMemo,
+	useState,
+} from 'react';
 import Select from 'react-select';
 import { NavLink } from 'react-router-dom';
 import { Alert, Button, Form, Tab, Tabs } from 'react-bootstrap';
 import { Trans, useTranslation } from 'react-i18next';
 import invert from 'lodash/invert';
-import map from 'lodash/map';
 import omit from 'lodash/omit';
+import zip from 'lodash/zip';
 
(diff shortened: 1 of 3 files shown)


www/src/Pages/PinMapping.tsx
```diff
@@ -63,6 +72,47 @@ const PinsForm = ({ savePins, pins, setPinAction }: PinsFormTypes) => {
...
+					onChange={(change) =>
+						setPinAction(
+							pin,
+							change?.value === undefined ? -10 : change.value, // On clear set to -10
+						)
+					}
+				/>
+			</div>
+		),
+		[],
```
andrepapoti, line 113: `options` technically is a dependency but there are no practical differences since it is a const

andrepapoti (commented):
LGTM

nit: You could use `useMemo` for `const options`

www/src/Pages/PinMapping.tsx
Pelsin: Thanks for checking the pr out!

I believe that options is not a dependency as it's not a reactive value, it's defined outside of the component

This even matching and exact example from the new react docs on how to use dependency array with non reactive value: 
From [react.dev](https://react.dev/)
```javascript
const serverUrl = 'https://localhost:1234';
const roomId = 'general';

function ChatRoom() {
  useEffect(() => {
    const connection = createConnection(serverUrl, roomId);
    connection.connect();
    return () => {
      connection.disconnect();
    };
  }, []); // ✅ All dependencies declared
  // ...
}
```

Source: https://react.dev/learn/lifecycle-of-reactive-effects#what-an-effect-with-empty-dependencies-means

arntsonl (approved):
Looks good!

PR #686, opened 2023-12-20
Added PS4 report speed hack to web config

arntsonl:
PS4 reporting is a hot topic.

This adds a switch to our web config so you can turn on a hack that disables report counters in the PS4/PS5 reports. This brings PS4/PS5 latency down to around 0.8ms, but could have potential instability in certain games.


arntsonl (commented):
Simple way to turn up the speed if we want to ignore report counters

PR #687, opened 2023-12-21
v0.7.6-RC2 Fix: Removed OLED and RGB focus features

arntsonl:
Remove the OLED and RGB focus features, they do not appear to work properly with focus mode.


arntsonl (commented):
Merging this in

Issue #688, opened 2023-12-21
Show current profile number on Display


hitsmaxft:
### Have you checked the GP2040-CE documentation [HERE](https://gp2040-ce.info/) to confirm this feature doesn't already exist?

Yes

### Which problem is this feature request solving?

I have three profiles for different scenarios: SFVI ranking/training and other arcade games. 
providing profile numbers on display would be more convenient for multiple button profiles players like me.

### Describe the solution you would like to see

show text on display to indicating current profile number 

profile 1(default ) : nothing or #1
profile 2 : #2
profile 3 or 4 :. #3 , #4

### Can you submit a pull request?

Yes

### Are you planning on working on this feature yourself?

Yes

PR #689, opened 2023-12-21
show profile number on display

hitsmaxft:
1. implementation of issue #688  show current profile number in display
2. append current firmware version in save filename

### DIFF
--- a/src/addons/i2cdisplay.cpp
+++ b/src/addons/i2cdisplay.cpp
@@ -1057,6 +1057,16 @@ void I2CDisplayAddon::drawStatusBar(Gamepad * gamepad)
 		case DPAD_MODE_LEFT_ANALOG:  statusBar += " L"; break;
 		case DPAD_MODE_RIGHT_ANALOG: statusBar += " R"; break;
 	}
+	switch (gamepad->getOptions().profileNumber)
+	{
+
+		case 2: statusBar += "#2"; break;
+		case 3: statusBar += "#3"; break;
+		case 4: statusBar += "#4"; break;
+		default:
+			statusBar += "#1";
+			break;
+	}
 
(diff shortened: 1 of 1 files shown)


mikepparks:
A few things of note here:
* Screen width is 21 characters wide. By adding "#n" after directional mode, the "macro" mode display is now being shifted off the screen by 1.
* Try not to mix contexts in your PRs. This should be two separate PRs since they're unrelated issues.

Please split the PRs into two, and consider another direction to display of profile number that doesn't impact the current status text.

hitsmaxft:
> A few things of note here:
> 
>     * Screen width is 21 characters wide. By adding "#n" after directional mode, the "macro" mode display is now being shifted off the screen by 1.
> 
>     * Try not to mix contexts in your PRs. This should be two separate PRs since they're unrelated issues.
> 
> 
> Please split the PRs into two, and consider another direction to display of profile number that doesn't impact the current status text.

i've remove backup file changes . 
about the text width , can i add a new option to  display config , to control this feature

src/addons/i2cdisplay.cpp
arntsonl, line 1: For this one, we will want a conditional for Macros since the 'M' will take the last character place. Once you add that, we can merge thank you!

arntsonl (changes requested):

src/addons/i2cdisplay.cpp
hitsmaxft: I've moved the profile number to last of status text and  make it disappear if profile number is 1 or macro enabled 
I think every work is done,  improvements would in next PR

arntsonl:
We have a major display rewrite coming soon so we're going to request a new PR once that has been completed. Lots of movement and dev on this thing right now, so we'll probably have a few conflicts like this that are not on the main branch.

@hitsmaxft once the display rewrite is done, we can look into displaying profile numbers!

PR #692, opened 2023-12-21
Update to default pico boardconfig.h and addons

TheTrainGoes:
This PR updates various addons to hold the defaults and not the individual boardconfig.h files.  This also cleans up the standard pico boardconfig.h


src/addons/turbo.cpp
```diff
@@ -8,6 +8,9 @@
 
 #include <algorithm>
 
+#define TURBO_ENABLED 0
```
arntsonl, line 11: You'll want to make sure these are not in here (turbo_enabled and turbo_led_pin) otherwise they are ALWAYS defined as these numbers. The #ifndef clause makes these come from boardconfig.h

src/addons/turbo.cpp
```diff
@@ -8,6 +8,9 @@
...
+
+#define TURBO_LED_PIN -1
```
arntsonl, line 13: Same with this one.

arntsonl (changes requested):

arntsonl (approved):
Looks good!

arntsonl (approved):

PR #695, opened 2023-12-22
Update to Flatbox Rev4 boardconfig.h

TheTrainGoes:
This cleans up the base boardconfig.h for the Flatbox Rev4 to a minimal state.

I have added in the switch numbers that are originally referenced


arntsonl (approved):
Much cleaner!

PR #696, opened 2023-12-22
Pin remapping in Peripheral page webconfig fixes

bsstephan:
This fixes issues with reclaiming or reserving pins on the Pin Mappings page when configuring peripherals on the Peripherals Mapping page.  This makes it so that pins can be set to unused if a peripheral (e.g. I2C0 for a disconnected display) has its pins set to "Unset", freeing up the pins for general usage on the Pin Mappings page. Vice versa is true, if pins are assigned to a peripheral, now they are properly reserved in the core pin mappings.

This also removes some deprecated code and marks the old peripheral pins, pre-manager, as deprecated.


arntsonl (approved):
Lots of good cleaning here!

PR #697, opened 2023-12-22
Add turbo rings to display for some layouts

TheTrainGoes:
This PR adds smaller rings to buttons on the OLED when turbo is active.

Currently this has only been done for the following layouts:
- SticklessB
- Vewlix8
- Noir8


arntsonl (approved):
I like these!

PR #701, opened 2023-12-24
Disable auth addons from enabling outside of their respected input modes

mikepparks:


arntsonl (approved):
Thank you for knocking this out so fast!

PR #702, opened 2023-12-24
fixbug: addons page submit not working while disabled addons validators could not passed

hitsmaxft:
when add-on is  disabled, validateRangeWhenValue and validateNumberWhenValue  still report error while config is wrong. 
but addons page never show errors, so looks like save button never works

this patch will ignore wrong configurations for disabled add-on by fix `is` condition


www/src/Pages/AddonsConfigPage.jsx
```diff
@@ -212,7 +212,8 @@ export default function AddonsConfigPage() {
 			onSubmit={onSuccess}
 			initialValues={defaultValues}
 		>
-			{({ handleSubmit, handleChange, values, errors, setFieldValue }) => (
+			{({ handleSubmit, handleChange, values, errors, setFieldValue }) => 
+			console.log('errors', errors) || (
```
hitsmaxft, line 216: show error to console log for developers

www/src/Contexts/AppContext.jsx
```diff
@@ -20,7 +20,7 @@ yup.addMethod(
 	'validateSelectionWhenValue',
 	function (name, choices) {
 		return this.when(name, {
-			is: (value) => !isNaN(parseInt(value)),
+			is: (value) => value > 0,
```
arntsonl, line 23: Instead of value > 0 which will throw a different error for non-numbers, let's make this two checks instead:

something like,

is: (value) => (!isNan(value) ? parseInt(value) : -1)

www/src/Contexts/AppContext.jsx
```diff
@@ -29,7 +29,7 @@ yup.addMethod(
 
 yup.addMethod(yup.NumberSchema, 'validateNumberWhenValue', function (name) {
 	return this.when(name, {
-		is: (value) => !isNaN(parseInt(value)),
+		is: (value) => value >0,
```
arntsonl, line 32: Same with these.

www/src/Pages/AddonsConfigPage.jsx
arntsonl: Makes sense, we don't have to worry about being too verbose!

arntsonl (changes requested):
I think our checks for isNaN and parseInt should be there, but just separate them out into two separate checks during one assignment

www/src/Contexts/AppContext.jsx
hitsmaxft: Currently "name" is all about the enabling state of addon, and they should be bound by the yup rule to be of type number. But for future code compatibility, I will modify and test it as you suggest.

www/src/Contexts/AppContext.jsx
```diff
@@ -15,12 +15,17 @@ yup.addMethod(yup.string, 'validateColor', function () {
 	);
 });
 
+const isEnabledValue = (value) => {
```
arntsonl, line 18: This will fail on arrays, [0,1] when it shouldn't. But you can flip it and it works fine, I'll take care of it and merge it in

arntsonl (approved):
Fixed up, ready to go

PR #708, opened 2023-12-26
Changed hold shortcut for switching to Bootsel mode to the right button (S1 + P1 + P2)

TenetDragon:
Changed hold shortcut for switching to Bootsel mode to the right button (S1 + P1 + P2)


TheTrainGoes (approved):
Looks good

PR #710, opened 2023-12-27
v0.7.6-RC2 FIX: Xbox One input-fix

arntsonl:
In order to increment sequence on key press in Xbox One, we have to keep track of last report sent.

This is slightly tricky because report generation does not guarantee a report is sent.
- So, if our last report sent input != current report sent, set sequence for this report to +1 but don't increment.
- ON SUCCESS, increment report by 1 and update our last sent report.


arntsonl (commented):
Merging this one in for RC3, we can hot-fix on RC3

PR #715, opened 2023-12-27
Removing info section that we can use backups from different versions

Pelsin:
Suggestion to remove this as we might not be able to guarantee backwards compatibility.


bsstephan:
Seconded.

arntsonl (approved):
Yeah makes sense. We might want different wording in the future but backups can be funky

PR #718, opened 2023-12-27
Remove locale en-gb as it was only made as a first example for the i18n

Pelsin:
Hopefully american english will suffice :)


SavageCore (approved):

arntsonl (approved):
Totally agreed, lets nuke it

PR #724, opened 2023-12-27
allow one GPIO pin to trigger 2+ buttons

bsstephan:
This adds a couple protobuf fields to the `GpioMappingInfo` structure to include dpad and button masks and uses them when setting up the `Gamepad` object to allow a pin to be the trigger for more than one button. All the cost is at setup time, so this shouldn't have any material effect on gamepad performance.

As usual, @Pelsin helped my proto-brain (pun intended) with the UI work. It will need improvement/expansion in the future, but this works for now.


bsstephan:
This works.

image

via

image

arntsonl:
Is this all ready to review?

arntsonl:
Confirmed this is working, going to have some other eyes on the UI as well.

Pelsin:
> Confirmed this is working, going to have some other eyes on the UI as well.

Thats good! Please ping me if the UI is not good enough for the first iteration, if so i will try to merge the new page with pin mapping page.

bsstephan:
Quick screenshot of the UI as it stands right now in this branch:
image

Note that this can only be done for the time being on the base mappings (profile 1).

arntsonl (approved):
Looks great! New changes we can put into PRs since this PR is already pretty huge

reimertz:
Thanks for this!

PR #726, opened 2023-12-27
v0.7.6-RC3 Fix: manage ASSIGNED_TO_ADDON profile pins as well as core pins

### LINKED ISSUE #667: Pin assigned to addon will not be unassigned
### Prerequisites

- [x] I am running the latest version of the firmware - built from latest main branch

### Context

* Firmware Version: latest main
* Name of device: pico
* Link to where you purchased the device: x
* Is this a custom build?: No
* Operating System: Windows
* Browser: Chrome

### Expected Behavior

Pins previously assigned to addon should be automatically unassigned if unused, but this is not always the case.

### Current Behavior

Taking Turbo addon for example
- If PIN14 was previously assigned as Turbo pin
- Then in the addon settings, change PIN14 to PIN16, and save
- Both PIN14 and PIN16 will show as "Assigned to addon" in the Pin Mapping page, even when PIN14 shouldn't be
- To unassign PIN14, you need to change the Turbo pin back to PIN14, save, then change to -1, save again

### Steps to Reproduce

As explained above.

### Screenshots & Files

x

bsstephan:
`docToPin` and our USB D+/D- handling code only managed the core profile configuration, but it also needs to set the fields in profiles.

fixes #667

### DIFF
--- a/src/configs/webconfig.cpp
+++ b/src/configs/webconfig.cpp
@@ -113,9 +113,13 @@ static void __attribute__((noinline)) docToPin(Pin_t& pin, const DynamicJsonDocu
 	{
 		pin = doc[key];
 		GpioMappingInfo* gpioMappings = Storage::getInstance().getGpioMappings().pins;
+		ProfileOptions& profiles = Storage::getInstance().getProfileOptions();
 		if (isValidPin(pin))
 		{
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Looks good!

PR #727, opened 2023-12-27
v0.7.6-RC3 FIX: Xbox One reporting issues

arntsonl:
Clean-up and two fixes:
- sequence no longer goes to 0
- keep alive does not trigger a repeat input report


arntsonl (commented):
Tested everything, holding buttons down while pressing guide, releasing buttons, letting keep alive occur in the middle of a press, we should be good!

PR #729, opened 2023-12-27
v0.7.6 Site Release

arntsonl:
Removed unneeded board assets, removed DURAL, PicoAnn, and AlpacaOwO from boards, updated cmake flow, changed SGF devices to SGF faust + bridget, added haute42


arntsonl (commented):
Updates for site

PR #732, opened 2023-12-28
single line regex to get current version in cmake script

hitsmaxft:
orgin cmake script's CMAKE_GIT_REPO_VERSION paring sometimes not working properly . here is my regex rewriting
do version in single regex  with regex grouping

example of output  from `cmake -B build` on my branch 
```
PICO_SDK_PATH is /home/wsl/projects/gp2040ce/pico-sdk
PICO platform is rp2040.
GIT_REPO_VERSION is v0.7.6-6-g40721a4
CMAKE_GIT_REPO_VERSION is 0.7.6
Submodules updated
```


bsstephan:
The goal of main's version is to produce v0.7.6.6 for your example. What error are you seeing?

hitsmaxft:
> The goal of main's version is to produce v0.7.6.6 for your example. What error are you seeing?
output fallbakc to 0.0.0
 I will
 upgrade regex to support 0.7.6.6

bsstephan:
That would be good. I have seen a surprising incidence rate of the 0.0.0 output so I can believe my successive hacks are causing a problem. If you get the regex to match to 0.7.6.6 in your example, this will probably be mergeable.

arntsonl (approved):
looks good, this catches all of the scenarios I can think of

PR #733, opened 2023-12-28
Fix 4-way mode when enabled for both core gamepad and DDI

### LINKED ISSUE #559: New 4-way implementation seems to conflict with itself when enabled for both normal and DDI directionals
Pasting from ~~IRC~~ Discord, as is my tradition:

> I tested more and it is stranger than I thought. bear with me:
> defining the desired behavior as Down -> Down+Right = Right and Right -> Down+Right = Down...
> 1) if 4-Way is enabled, the desired behavior is present on the main directionals
> 2) if Dual Directional 4-Way is enabled *and normal 4-Way is disabled*, the desired behavior is present on the secondary directionals
> 3) if *both* 4-Ways are enabled, the behavior on both changes to Down -> Down+Right = Right and Right -> Down+Right = Right, giving preference to left/right (not up/down as I said initially)
> I don't immediately have an explanation for why this is, but my guess is something with the history is getting into a bad state if both sets of directions are contributing to the history?
> this is probably a very strange corner case and maybe only meaningful to someone like me testing a bunch of stuff after code changes

Writing this down for when myself or @sirrow can take a look.

bsstephan:
This fixes the misbehavior described in #559 item 3, where 4-way didn't work properly if it was enabled for both the core gamepad and DDI at the same time. @sirrow is the author, I have just pulled, reviewed, and tested it.

Fixes #559.

### DIFF
--- a/headers/addons/dualdirectional.h
+++ b/headers/addons/dualdirectional.h
@@ -56,6 +56,8 @@ public:
 private:
     void debounce();
     uint8_t gpadToBinary(DpadMode, GamepadState);
+    uint8_t updateDpadDDI(uint8_t dpad, DpadDirection direction);
+    uint8_t filterToFourWayModeDDI(uint8_t dpad);
     void SOCDDualClean(SOCDMode);
     uint8_t SOCDCombine(SOCDMode, uint8_t);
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Ahh interesting, that's a weird bug but looks good

PR #734, opened 2023-12-28
Update documentation to reflect the pin mapping webconfig changes

bsstephan:
This updates the documentation for:

* Extra Button (docs removed entirely, note added pointing at Pin Mappings)
* Dual Directional Input (docs updated)
* Joystick Selection Slider (docs updated)
* SOCD Selection Slider (docs updated)

...to remove references to the pins that are now managed at Pin Mappings. Screenshots updated and examples of the Pin Mappings entries provided.

This also fixes a tiny issue in the developer docs, I'm assuming the docs were moved into `site/` at some point and the directions not updated.


arntsonl (approved):
Nice, yeah we probably have some documentation debt to catch up on

PR #735, opened 2023-12-28
v0.7.6 FIX: deduplicate the PLED PWM GPIOs vs. RGB indexes

bsstephan:
I missed/lost track of the fact that pledPin1-4 are only true GPIOs in PWM mode, and RGB mode uses the same fields as generic indexes instead. this is of course a wreck for tracking in GpioMappings, so I added new pledIndex1-4 fields for use in RGB mode, letting pledPin1-4 be truly just for GPIO pin assignments in PWM mode (or set to -1 otherwise).

furthering the confusion, the web UI *already* referred to pledIndex1-4, just it copied the value of those states back and forth to/from pledPin1-4, so now that the API has those fields as real fields, I just had to remove the copying.

I don't have PLEDs so all I was able to test is that protobuf does what it should now, and only when PWM pins are set do they get reserved in GpioMappings. tested on a nuked board, with webconfig allowing for getting out of a bad state.

### DIFF
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -259,6 +259,11 @@ message LEDOptions
 	optional uint32 pledColor = 30;
 
 	optional bool turnOffWhenSuspended = 31;
+
+	optional int32 pledIndex1 = 32;
+	optional int32 pledIndex2 = 33;
+	optional int32 pledIndex3 = 34;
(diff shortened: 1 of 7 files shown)


arntsonl (approved):
No worries, a lot of moving pieces!

PR #737, opened 2023-12-28
Add backup route to spa paths in webconfig, remove extra spaces

Pelsin:
Solves the issue of not being able to navigate to /backup directly


arntsonl (approved):
Oh nice! we were looking for this

PR #738, opened 2023-12-29
v0.7.6 FIX: Changed construction of peripheral devices to not require "new"

mikepparks:


arntsonl (approved):
Yeah heap vs. static stack is such a weird one on rp2040. Our core0/core1 memory sharing is a total hack but it works :)

PR #752, opened 2024-01-02
Validate addon settings when enabled

Pelsin:
Solves the issue of the addons page thinking all addons are enabled even when disabled.
Current we have this issue: 
Skärmavbild 2024-01-02 kl  00 40 03

The `is` check is always compared against the [checkbox/toggle](https://github.com/OpenStickCommunity/GP2040-CE/blob/aa48d716c654b55da66d0e955576d3844c832e5d/www/src/Pages/AddonsConfigPage.jsx#L204) value for the addon section, meaning it's either `1` or `0`
Yup validation for `is` does a strict equal hence not being able to do `is: true`

This change is equivalent to 
```javascript
is: (value) => value === 1
```
If we feel that is more explicit

More info: https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema

### DIFF
--- a/www/src/Contexts/AppContext.jsx
+++ b/www/src/Contexts/AppContext.jsx
@@ -15,16 +15,12 @@ yup.addMethod(yup.string, 'validateColor', function () {
 	);
 });
 
-const isEnabledValue = (value) => {
-	return Number.isInteger(value) && value > -1;
-}
-
 yup.addMethod(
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Much cleaner!

PR #757, opened 2024-01-02
Update Haute42 boardconfig.h

TheTrainGoes:
This update changes the default max LED brightness to be 100 (up from 50).  This matches the original brightness from their backup file in v0.7.5.

This update also changes the default LED animation to solid fade.   This is also more in line with their backup file in v0.7.5 and works better on the unit given the LED order.


arntsonl (approved):
Fresh & clean!

PR #765, opened 2024-01-03
Update to i2cdisplay defined values

TheTrainGoes:
This PR updates i2cdisplay.h's `HAS_I2C_DISPLAY` to 0 which was current set as -1 and causing issues.

This was only discovered now that we are stripping unused instances of `HAS_I2C_DISPLAY` out of configs where not used.


arntsonl (approved):
Awesome, hard to find bug!

PR #766, opened 2024-01-03
Regex Hash & Hash-Dirty 0.0.0 Replacement

arntsonl:
Catch both hash values and hash values with a flag set such as -dirty


arntsonl (commented):
Works for all the folks having issues

PR #767, opened 2024-01-03
Updated webconfig.cpp to fix flipping display bug

TheTrainGoes:
This update fixes a bug in webconfig.cpp which did not allow you to `mirror` or `mirror + flip` a display.  This is because originally we only had two values (`none` and `flip`) and they were set in webconfig.cpp as 0-1.  That has been removed and now all four modes can be used.


arntsonl (approved):
Good catch!

PR #774, opened 2024-01-03
Update the Mavercade Keebrawler boardconfig

TheTrainGoes:


arntsonl (approved):
Looks good, makes sense so we can support more Mavercade stuff

PR #776, opened 2024-01-03
Initialize peripherals on both cores to prevent lockups in webconfig

mikepparks:


arntsonl (approved):
Yep, this makes sense. Core0 and Core1 cross-chatter seems very "not tested" on the Pico-SDK, I'm guessing we do things they never thought you could run from one rp2040.

PR #777, opened 2024-01-04
Further cleanup of all boardconfig files

TheTrainGoes:
This PR removes all instances of defines in the boardconfig.h files that are already at the default value for that define.


arntsonl (approved):
Lots of QoL improvements!

PR #786, opened 2024-01-09
Add turbo rings for Fightboard right screen display

TheTrainGoes:
This PR adds turbo rings to the fightboard right display


arntsonl (approved):
I like it!

PR #791, opened 2024-01-11
fix issue 790

### LINKED ISSUE #790: JS Selection Slider debouncer somehow causes LS (default) and RS (pin mapping) to oscillate at debounce time
> so if I have default let to LS, buttons set to DP and RS, if I hold the DP button the lever always registers DP (as expected), if I hold the RS button the lever seemingly randomly oscillates between registering LS and RS (I say randomly because it's not entirely even, I think it's picking RS most of the time)
> this is the same as what IA reported
> if I recompile with `#define JSLIDER_DEBOUNCE_MILLIS 0`, no oscillation, RS button works as it should
> so something in the addon's debouncer is acting strange
> going to try a huge value and see what happens
> yup
> if I set it to 250ms, I see LS activating far less in my tester
> I'm only eyeballing it, but I could believe it's 4 times a second
> so something in the debouncer is momentarily setting the mode back to LS
> I'm in the middle of trying another thing (and I guess figuring out why my webconfig is crashy, so two things), so I can't look at that code right now, but I would look at the bitwise operations inside `if ((uNowTime - uDebTime) > JSLIDER_DEBOUNCE_MILLIS)`

NickGuyver:
Fix for issue 790 where regression from 604 missed breaks to match original if-else.

### DIFF
--- a/src/addons/jslider.cpp
+++ b/src/addons/jslider.cpp
@@ -65,10 +65,12 @@ void JSliderInput::debounce()
             // Bounce Right Analog
             case DPAD_MODE_RIGHT_ANALOG:
                 dDebState = (DpadMode)(dDebState ^ DPAD_MODE_RIGHT_ANALOG);
+                break;
 
             // Bounce Left Analog
             case DPAD_MODE_LEFT_ANALOG:
                 dDebState = (DpadMode)(dDebState ^ DPAD_MODE_LEFT_ANALOG);
+                break;
         }
         uDebTime = uNowTime;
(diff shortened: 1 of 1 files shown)


bsstephan:
I think this code looks good, but I seem to still have the issue even having built with it locally. I will look more when I get a chance.

bsstephan:
It seems like #793 is the fix even if we don't understand it, and that gets rid of the switch, so we don't need this one anymore. Thanks for the help in looking at this though @NickGuyver !

PR #792, opened 2024-01-11
PS4 latency improvement: force occasional reports (with new counters) in PS4 mode to reduce the need for the PS4 Hack

bsstephan:
~~This does some trickery with the report increments to use them to force a PS4 report roughly every 7ms (there's no particular design behind this value, it's just the result after I looked at some docs)~~ This puts a forced report with new counter values on a ~~17~~ 1000 ms timer, which fixes games that expect occasional reports, meaning we may not need the PS4 Hack anymore.

Shoutout to @TheTrainGoes for doing some quick latency testing. See the last comment for updated timer-based latency results, they are better than the old methodology anyway.

The latest results are virtually no increase over the measured 0.86 ms (in the README) for PS4 Hack mode, and considerably better than the 1.72 ms measured without the hack. A game thinking the inputs are stuck will have a new report every second to potentially correct itself, so this is converging on a sweet spot of game accuracy + low latency.

### DIFF
--- a/src/gamepad.cpp
+++ b/src/gamepad.cpp
@@ -974,7 +974,8 @@ PS4Report *Gamepad::getPS4Report()
 	if ( !options.ps4ReportHack ) {
 		uint32_t now = to_ms_since_boot(get_absolute_time());
 		if ((now - keep_alive_timer) > PS4_KEEPALIVE_TIMER) {
-			ps4Report.report_counter = last_report_counter++;	// report counter is 6 bits
+			last_report_counter = (last_report_counter+1) & 0x3F;
+			ps4Report.report_counter = last_report_counter;		// report counter is 6 bits
 			ps4Report.axis_timing = now;		 		// axis counter is 16 bits
 			keep_alive_timer = now;
 		}
--- a/src/gamepad.cpp
+++ b/src/gamepad.cpp
@@ -21,6 +21,9 @@
 // Xbox One compatibility
 #include "xbone_driver.h"
 
+// number of report cycles before forcing a new report --- approx. 7ms (see usage below)
+#define PS4_REPORT_DELAY_MAGIC		188
+
 // MUST BE DEFINED for mpgs
 uint32_t getMillis() {
 	return to_ms_since_boot(get_absolute_time());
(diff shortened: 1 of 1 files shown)


bsstephan:
As discussed, I'm taking a timer-based approach and will add the PS4 Hack back to do A/B testing, but I'm hoping the upcoming change will be as good if not better, and we can eventually move off of the hack.

bsstephan:
New numbers from @TheTrainGoes against the 17ms timer method:

## Hack off

```
min: 0.55
max: 2.12
average: 0.86
stddev: 0.27

min: 0.55
max: 2.24
average: 0.88
stddev: 0.27

min: 0.55
max: 2.05
average: 0.87
stddev: 0.26

min: 0.55
max: 2.24
average: 0.88
stddev: 0.29
```

## Hack on

```
min: 0.55
max: 1.33
average: 0.84
stddev: 0.24

min: 0.55
max: 1.38
average: 0.86
stddev: 0.24

min: 0.55
max: 1.38
average: 0.86
stddev: 0.24

min: 0.55
max: 1.26
average: 0.86
stddev: 0.24
```

bsstephan:
5000 ms timeout takes longer for Ultimate Chicken Horse to correct itself, but latency is better:

```
min: 0.55
max: 1.36
average: 0.85
stddev: 0.24

min: 0.55
max: 1.34
average: 0.86
stddev: 0.24

min: 0.55
max: 1.38
average: 0.85
stddev: 0.24

min: 0.55
max: 1.32
average: 0.88
stddev: 0.24
```

bsstephan:
1000 ms is the same, essentially.

## Hack off

```
min: 0.55
max: 1.37
average: 0.85
stddev: 0.24

min: 0.55
max: 1.38
average: 0.85
stddev: 0.24

min: 0.55
max: 1.79
average: 0.87
stddev: 0.24

min: 0.55
max: 1.37
average: 0.86
stddev: 0.24
```

Thanks again to @TheTrainGoes for running my builds to get preliminary numbers.

arntsonl (approved):
Looks good! Reduces our non-hack version of PS4 reporting and unlocks menu-screen locks when they occur.

PR #793, opened 2024-01-12
use if-else block logic instead of switch in jslider debouncer

### LINKED ISSUE #790: JS Selection Slider debouncer somehow causes LS (default) and RS (pin mapping) to oscillate at debounce time

bsstephan:
I'm not going to say we *understand* why this fixes #790, but it appears to.

### DIFF
--- a/src/addons/jslider.cpp
+++ b/src/addons/jslider.cpp
@@ -58,21 +58,14 @@ void JSliderInput::debounce()
     }
 
     uint32_t uNowTime = getMillis();
-
     if ((uNowTime - uDebTime) > JSLIDER_DEBOUNCE_MILLIS) {
-        switch (dpadState ^ dDebState)
-        {
-            // Bounce Right Analog
-            case DPAD_MODE_RIGHT_ANALOG:
-                dDebState = (DpadMode)(dDebState ^ DPAD_MODE_RIGHT_ANALOG);
-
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Such weird optimizations in ARM GCC

arntsonl:
@jfedor2 nailed it, the switch statements didn't have breaks in them!!

PR #794, opened 2024-01-14
Adds a "pin viewer" in pin mapping page

Pelsin:
Also moved info text to the top and fixed broken translation resulting in a link to the reset page to not be visible

https://github.com/OpenStickCommunity/GP2040-CE/assets/5345892/c24eab22-a820-4638-9fee-a84f70a1012c


arntsonl (approved):
This is a really neat utility function to our pin mapping page!

PR #795, opened 2024-01-14
Make sure the action button labels update when changing mode

Pelsin:
https://github.com/OpenStickCommunity/GP2040-CE/assets/5345892/f3ecccaa-8b86-4d01-ab4b-ca4114ee49a8

Solves a regression bug from the pin rework, issue found by thetrain


arntsonl (approved):
This might not fix the immediate issue but this is something we need

PR #796, opened 2024-01-14
GPIO PinXY Assignment in Board Config

arntsonl:
All pin assignments for board config are now defined as:

#define GPIO_PIN_XY GpioAction::BUTTON_XY

I also migrated over the Turbo button to our pins profile, and fixed a few spots where the defaults were set to old board config assignment.

Display will show TXY in the OLED if Turbo is on now instead of validating the turbo button, but we can revisit this.


arntsonl (commented):
Double checked everything

PR #800, opened 2024-01-16
Updated copyrights to 2024, moved copyrights for MIT over from feral …

arntsonl:
Moved copyrights for MIT over from feral to OpenStickCommunity for files we know we created (legacy from copy-paste of old project).


arntsonl (commented):
Everything is up to 2024 that should be

PR #803, opened 2024-01-17
Add .elf to artifacts and run on pull_request

jfedor2:


arntsonl:
I'm okay with this as a single elf for testing purposes. Also good catch on build report, we still double build for some reason but we'll figure it out

PR #811, opened 2024-01-20
Fix for Xbox One / USB Host Soft-Reboot

arntsonl:
This fixes the first-boot and reboot issues we were seeing on Xbox One with the dongle plugged in. We have to tell TinyUSB Host to shut down the bus/port before we perform the soft reeboot.

Added a shutdown() feature to USB Host Manager to take care of Controller -> Web Config -> Controller bad behavior.


arntsonl:
Tested and verified this was the issue. This also solves first-boot on Xbox One not working.

PR #813, opened 2024-01-21
LED button press fade across all themes. 

jmrochuk:
Move LED animation fade into animation base class so it works for all themes
Move fade up/down shortcut to S1+S2 + L3/R3


arntsonl (approved):
Looks great, super clean! Really good improvements on moving around those variables.

This was also tested a bunch at my locals and everyone loved it!

PR #815, opened 2024-01-22
Update the Open_Core0 images

TheTrainGoes:
This RP updates the Open_Core0 images as well as provides a button layout, pin mapping and RGB order


arntsonl (approved):
Looks good to me!

Issue #826, opened 2024-01-29
Input history shows locked buttons being pressed when focus mode is turned on


jnkhac:
### Prerequisites

Please check the following before posting an issue / bug report.

- [ ] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: v0.7.7 Custom Compiled
* Name of device: Fightingbox Flatbox Ultra
* Link to where you purchased the device: https://www.aliexpress.com/item/1005006286226142.html
* Is this a custom build?: No
* Operating System: Windows 10
* Browser: Firefox

### Expected Behavior

When focus mode and input history is enabled, flipping the focus mode switch to the on side should no longer register any of the locked buttons (usually S1, S2, L3, R3, A1, A2) in the input history.

### Current Behavior

Locked buttons when focus mode is on still appears in the input history when pressed. The inputs registered in the history also looks like it is being repeatedly pressed.

### Steps to Reproduce

1. Focus mode and input history enabled (with default settings)
2. Flip the focus mode switch on controller to the on side
3. Pressed any of the locked buttons 
4. Check if it appears in the input history

### Other Notes

* Focus mode functionally works fine, issue seems to come from input history
* May only be an issue on devices with a focus switch (e.g. open_core0)
* May only be affecting my device and I don't have any other RP2040 based devices to test issue
* I can't upload my BoardConfig.h file but I can copy paste the code later in a comment if needed

TheTrainGoes:
This has been fixed as of v0.7.11 and this issue will be closed.

Tested on my Haute42COSMOX C16 with the turbo button remapped to be focus.  Focus setup with S1/2 A1/2 blocked out on hold.  Holding the remapped focus button no longer shows blocked inputs on the input history.

PR #827, opened 2024-01-29
fix input history showing locked buttons for issue #826

jnkhac:
I'm not sure if issue #826 affects any other devices but the changes I made here seems to work on my device.


TheTrainGoes:
Hey @jnkhac - I tested this but it was not working as intended.  

I will test again tomorrow on it.

TheTrainGoes:
Hi @jnkhac ,

Following up to see if you had any updates for this.

jnkhac:
Hi @TheTrainGoes,

Sorry I've been putting this off for a bit. I haven't been able to find any comments on the discord about the issue so I'm guessing that it only happen on my controller. I've also tried the firmware from the seller and the same issue still occurs. I thought it might've been a hardware issue so I opened up my controller to see but I don't know enough to tell if there's a problem other than looking at the soldering job.

Could you tell me what controllers you tested this with? It will just be for future reference when I buy another controller to test it on. Other than that I'll close the PR in a few days since the fix doesn't seem to work on all controllers and I don't have another controller to test it currently.

PR #830, opened 2024-02-04
GP2040-CE Input Driver system rehaul

arntsonl:
This is a major overhaul of the USB input driver system.

All input drivers have been moved to their own child class with GPDriver set as the parent class interface.

Why: This makes adding new systems, modifying current systems, and maintaining support WAY easier.

How: This greatly simplifies where each report, report size, update mechanism, and USB calls lives and keeps the majority of our TinyUSB host code intact.

This also moves the update, send_report, send_report_cb, and get_report logic over to the input driver to handle all of these changes. this will give the input driver more control over USB and let gp2040 main pass this along.


arntsonl (commented):
Everything is looking clean

Issue #785, opened 2024-01-06
PS3 port of MvC2 doesn't work on our firmware (as of 0.7.6)


bsstephan:
Reported by @TheTrainGoes and tested by me. The Darkstalkers and Jojo's ports work fine, so (so far) seems specific to MvC2 somehow.

I was going to get a PS3 TE1 out of storage at some point and confirm that it works. I was originally thinking I was going to be sifting through a bunch of reports to find differences and try them out, but come to think of it from another conversation, maybe I should just try hacking the VID/PID first and see if that gets us through to the game.

bsstephan:
I suppose I should qualify "work", but it's basically any form of button functionality. Turning on LS emulation would let you navigate the top level (e.g. outside of game emulation) main menu, but even so, none of the buttons activated any menus. DPad also didn't do anything. Could still get to the XMB via the Home button, IIRC, and other games worked fine with our firmware before and after, so the controller was fine, just MvC2 didn't care.

TheTrainGoes:
To add further to this - has been an issue for a long time.  Although you may want to check v0.4.3 of the original GP2040 to see if it is still an issue there.  On the flip side the home button did not work on GP2040.

bsstephan:
Further discussions between us have us collectively curious/interested in if we can get this to work (perhaps justifying a PS3 mode with quirks separate from generic USB HID), so I'm reopening it so it's on my visible list.

PR #838, opened 2024-02-15
remove some of the user-facing references to "DInput"

bsstephan:
I wrote this for myself and couldn't help making a PR. sorry not sorry


bsstephan:
We got a spirited conversation out of this PR, so mission accomplished, but we're agreed to close this for now. Thoughts/threads, for future reference/consideration:

* do users recognize DInput more than they recognize USB/USB HID/HID? if we changed it, are they confused or is it just an opaque label to them?
* does the average PC user care at all, and what they really need is simply "PS3", because they're probably using XInput on PC?
* should PS3 mode be different than USB HID mode? (see #785 --- does MvC2 expect specific quirks that would justify a separate mode?)
* PS3 mode as HID with quirks would match all of the other non-XInput consoles, which are *also* essentially "HID with quirks" modes
* if we *did* need PS3-as-HID-with-quirks as a separate mode, would/could we then have generic HID do crazier stuff, like configure more axes or buttons or have extra sub interfaces or ...

I continue to live on my hill but concede these points/conversations to later. Maybe resurrected depending on the outcome (if any) of #785.

PR #840, opened 2024-02-17
populate the PS3 analog values for buttons

### LINKED ISSUE #785: PS3 port of MvC2 doesn't work on our firmware (as of 0.7.6)

bsstephan:
PS3 face buttons are technically analogs like L2/R2. they were declared in the payload but never used, but apparently Marvel vs. Capcom 2 uses the analog values, so use the digital button state to populate the fake "analog" values and get the game working.

this also confirms the order of the axes declared in HIDDescriptors.h, which is nice.

this *ALSO* means MvC2 (and who knows what else) has a predefined notion of what's in the payload bytes despite the interface descriptor, so that's interesting, but whatever

I WANNA TAKE YOU FOR A RIDE

Fixes #785

### DIFF
--- a/headers/drivers/hid/HIDDescriptors.h
+++ b/headers/drivers/hid/HIDDescriptors.h
@@ -102,8 +102,7 @@ typedef struct __attribute((packed, aligned(1)))
 	uint8_t r_x_axis;
 	uint8_t r_y_axis;
 
-	// Gonna assume these are button analog values for the d-pad.
-	// NOTE: NOT EVEN SURE THIS IS RIGHT, OR IN THE CORRECT ORDER
+	// button analog values for the d-pad.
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
looks good! doesn't change anything on PC and that's great it fixes PS3/MvC2

PR #841, opened 2024-02-17
ADS1256 SPI analog addon

FeralAI:
This PR is for a new addon for the ADS1256, a low noise 24-bit 8-channel Analog-to-Digital converter that communicates over SPI. This addon is modeled after the existing ADS1219 addon, so extra features like deadzone config and autocalibration are not present. This work includes:

* Implemented SPI driver in PicoPeripherals library, including support for DMA (not used for ADS1256, but there when needed)
* Ported [this ADS1256 library](https://github.com/CuriousScientist0/ADS1256) into GP2040-CE and uses the PicoPeripherals SPI driver
* Fixed support for `FLOAT` and `DOUBLE` in the Protobuf implementation. The macros needed to support those data types were not present in `config_utils.cpp`.
* Added ADS1256 section to the Addons page

Screenshot 2024-02-16-211200

For the addons page, SPI pulls from the configured peripherals and the addon will not allow configuration if at least one SPI block is not configured, similar to the ADS1219 addon. It will also present a dropdown to select the CS pin using a combination of the selected SPI block's HW CS pin and other available pins.


mikepparks (dismissed):
Code checks out on my end. @FeralAI tested enough on hardware that implementation testing is good enough for me.

arntsonl:
Nice, I'll do a once over but this is looking clean to me

mikepparks (approved):

arntsonl (approved):
Looks great! Thank you @FeralAI !!!

PR #842, opened 2024-02-17
Stress - Switch incorrect pin 3 for pin 2

SavageCore:
Putting this here to build it, but it should work!


SavageCore:
Works fine!

arntsonl (approved):
Yep, this was a pin mistake on stress board! Thank you @SavageCore !

PR #845, opened 2024-02-18
Added button masks for Neo Geo ASP extended buttons

### LINKED ISSUE #844: Adding Support for Neo Geo Arcade Stick Pro emulation so 6 buttons can be used instead of 4 for 2 players on Neo geo Arcade Stick Pro
### Have you checked the GP2040-CE documentation [HERE](https://gp2040-ce.info/) to confirm this feature doesn't already exist?

Yes / Partialy works with neo geo mini mode 

Im able to use neo geo mini controller mode plugged into an Arcade Stick pro  but its limited too 4 buttons and 6 are achivable if another Arcadestick pro is connected to it so id like to be able to emulate an Arcadestick pro controller in neo geo/pc mode 

I think that adding ASP support as an option will help many people improve their experience with the console as it is a great option too play games but lacks a cheap way to add a 2nd player with full button support 

they say mayflash has it available in an update but im not sure about it 

/No. I dont understand what this means 

I dont know how to do it myself or i would do it and share it but  i would gladly help gather any info from the device thats needed and would love to test as i have many (5) devices using gp2040-ce i could do testing with

mikepparks:
This should allow the side panel Options button and 4 extra action buttons to be included where used.

Assume the following mapping, as the remaining 4 buttons are unlabeled (using L1/L2/R1/R2 naming).
```
A C R1 L1
B D R2 L2

B3=A
B1=B
B4=C
B2=D
S1=Select
S2=Start
A1=Options
L1=L1
L2=L2
R1=R1
R2=R2
```
This should help to resolve #844 .

### DIFF
--- a/headers/drivers/neogeo/NeoGeoDescriptors.h
+++ b/headers/drivers/neogeo/NeoGeoDescriptors.h
@@ -23,11 +23,15 @@
 // Button report (16 bits)
 #define NEOGEO_MASK_A       (1U <<  0)
 #define NEOGEO_MASK_B       (1U <<  1)
+#define NEOGEO_MASK_R1      (1U <<  2)
 #define NEOGEO_MASK_C       (1U <<  3)
 #define NEOGEO_MASK_D       (1U <<  4)
+#define NEOGEO_MASK_R2      (1U <<  5)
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Looks good! Works perfect on my PC

PR #847, opened 2024-02-19
Buzzer Speaker off-by-one fix

arntsonl:
Buzzer speaker had an off-by-one issue. So the last note would read from memory, and in some cases cause a "fart" noise


arntsonl:
Confirmed working. I was able to replicate the off-by-one on the web config boot up jingle (extra note added).

PR #848, opened 2024-02-21
PNG Reduction

arntsonl:
Testing our uf2 size with a small png reduction and removing an unused asset


arntsonl:
Woo 4.1kb shaved off for free!

PR #854, opened 2024-02-22
Debounce delay fix

arntsonl:
Debounce delay should be a 32-bit rather than a 16-bit


arntsonl:
The other solutions are looking much more promising!

PR #856, opened 2024-02-23
Tidy up singleton pattern used in Peripheral Manager

MajorArkwolf:
It appears that the peripheral manager had been converted to use a modern cpp singleton pattern but had a few small issues that this PR addresses

1. The constructor (and implicitly the destructor) was not private, this would allow a developer to construct additional managers which could lead to erroneous behaviour. 
2. The copy assignment was returning void, while it is fine, T& is the recommended return type. (https://en.cppreference.com/w/cpp/language/copy_assignment)

### DIFF
--- a/headers/peripheralmanager.h
+++ b/headers/peripheralmanager.h
@@ -9,9 +9,8 @@
 
 class PeripheralManager {
 public:
-    PeripheralManager(){}
 	PeripheralManager(PeripheralManager const&) = delete;
-	void operator=(PeripheralManager const&)  = delete;
+	PeripheralManager& operator=(PeripheralManager const&)  = delete;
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Looks good

arntsonl (changes requested):
The code does not compile, please address thank you!

mikepparks:
I had initially made these private, but the exact error had come up for me. I would say that going forward, please make sure your code compiles before submitting a PR.

arntsonl:
This is actually a two part problem, there was an erroneous PeripheralManager in the gp2040 class trying to call a constructor. I'm going to close this PR and generate a new one with the correct fix.

PR #857, opened 2024-02-23
Pin debouncer for buttons rather than gamepad debouncer + improved logic

bsstephan:
A couple classes had their own debouncer or reference to the gamepad debouncer. this seemed inefficient and caused potential for incongruent behavior, thinking that instead we should just debounce the GPIO pins (for buttons). this does that.

Meanwhile, Feral questioned the logic of our gamepad debouncer as of 0.7.6, which was a plausible explanation for reports of double inputs cropping up lately. Feral's bugfix has been incorporated into this refactor and we're reaching out to see if it addresses those issues.


arntsonl (approved):
Fantastic work, thank you @bsstephan . This is so much cleaner, one centralized place to handle our debouncing logic.

strusic:
Ive tried this fw, and debouncing works fantastic, but on xbox mode X button wont respond, sometimes I got shutdown menu on my xbox. On oled screen it show that button is pressed.

arntsonl:
@strusic can you make an issue for that? I'll look into it, its probably something I didn't catch with the logic I'm using to do xbox one. 

https://github.com/OpenStickCommunity/GP2040-CE/issues

strusic:
@arntsonl done, as you requested. Looking forward to new pull that fix this ;)

PR #858, opened 2024-02-23
Corrected fix for PeripheralManager as a singleton

arntsonl:
This is the correct fix for PeripheralManager as a singleton


MajorArkwolf:
You missed privatising the destructor and fixing the return type on the copy assignment.

PR #859, opened 2024-02-24
Tidy up last of Perpherial Manager

MajorArkwolf:
- Made destructor explicitly private
- Fix return type on copy constructor
- Remove comment about thread safety (As far as I can tell, there is nothing here that guarantees thread safety. A singleton just gives the variable a static location but its still prone to race conditions)

### DIFF
--- a/headers/peripheralmanager.h
+++ b/headers/peripheralmanager.h
@@ -9,9 +9,8 @@
 
 class PeripheralManager {
 public:
-    PeripheralManager(){}
 	PeripheralManager(PeripheralManager const&) = delete;
-	void operator=(PeripheralManager const&)  = delete;
+	PeripheralManager& operator=(PeripheralManager const&)  = delete;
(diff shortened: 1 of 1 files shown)


MajorArkwolf:
@arntsonl @mikepparks Sorry I was asleep when you guys reviewed my initial PR. The CI workflow was failing on the main from an unrelated change, so I figured I would wait until the morning to compile it locally (https://github.com/MajorArkwolf/GP2040-CE/actions/runs/8019345772/job/21907051731). 

Thanks for having time to review though, much appreciated.

MajorArkwolf:
Okay confirmed this compiles locally, moving out of draft PR.

arntsonl:
This does not match our coding on other singletons. Please submit PRs with feature changes and meaningful updates and use our Discord server to discuss what everyone wants to see in the firmware. 2-3 line code clean-ups have less value unless they fix a known issue. The singleton fix I did was an actual issue as we had an extra instance inside of gp2040 which just takes up RAM.

PR #868, opened 2024-02-27
PS4 de-hack: only force a report when the reports are otherwise idle

bsstephan:
Instead of sending a forced report every X ms no matter what, this moves the forced report to only occur when the new report matches the last sent report for X ms, since if the user is creating lots of valuable (changing) input, we don't need to force one.

I think 250ms is fair in this new model, but we can test it later and bump it up to say 500ms if we don't like 250ms for whatever reason. 250ms seemed like a good compromise between stuck inputs and minimal effect on active gaming, but we can dial it in.


arntsonl (approved):
Looks perfect, this is a much more eloquent fix to a strange problem.

PR #879, opened 2024-03-04
Dual rotary encoder support

mikepparks:
Added early support for dual quadrature rotary encoders
* Currently only maps to Gamepad-related items (analog sticks and triggers + d-pad) and keeps those values within range (no overflowing)
* The pulses per revolution can be customized, but currently untested beyond 24PPR


arntsonl (changes requested):
Looks great! Works fine on my testing with a 20 PPR, 600 PPR is pending as we are not sure if its possible yet.

arntsonl (approved):
Looks great

PR #881, opened 2024-03-06
Buzzer Speaker enable pin

mikepparks:
Added optional enable pin to to allow toggling the sound on/off when necessary. The enable pin is pulled high on setup() if the pin is defined so it will need to be brought low to disable the speaker during runtime.


arntsonl (approved):
We'll need to update the docs on this, but our 0.001% of users that use buzzer will figure it out :)

PR #883, opened 2024-03-08
Fix the Turbo LED state

SavageCore:
The logic didn't match the comments above it or what Feral, Train and I discussed the other day on Discord.

This is now working right, it seems to behave how I'd expect on my stick.


arntsonl:
Replaced by PR#898

PR #886, opened 2024-03-09
Communication between GP2040-CE instances

FeralAI:
This draft is a WIP for communication between two GP2040-CE devices. The structure is in place to support I2C and SPI, and can easily be expanded to UART or some other protocol.

Only I2C gamepad state is functional at this time, but seemed pretty stable in the bit of testing I've done so far. I'm working through some timing issues with the SPI side, but will run some latency tests using our current methods for comparision once SPI is ready.

Next I'm going to add the config UI, which will be two options on the Peripheral Mapping page that reflect the GPCommsOptions object in the config proto file.

After we discuss and verify the approach, I'll create a new branch from this work for a cleaner history instead of bringing in my nonsense 😁


headers/storagemanager.h
FeralAI, line 1: Revert changes, no need since helper function was removed.

PR #891, opened 2024-03-14
Adds button on each profile mapping to copy from base profile

Pelsin:
Screenshot 2024-03-14 at 18 38 19

Still requires saving after copying, copy/design suggestions or is this ok?


arntsonl (approved):
I like this! Feature works great, makes it simple to copy the base profile over to other profiles and language works well.

PR #892, opened 2024-03-15
New Driver System - Fix bug for Wii/Tilt/few others

arntsonl:
This should fix issues with state updates before the driver has been declared. We might want to revisit assigning mid to a Gamepad variable instead of pulling from the driver input, but this will fix crashing issues for now.


arntsonl (commented):
Verified and fixed

PR #893, opened 2024-03-15
unassign the old addon pin after assigning the new one

### LINKED ISSUE #816: Addon pin assignment not released after changing via webconfig
### Prerequisites

### Context

* Firmware Version: 0.7.7
* Name of device: OpenCore0
* Link to where you purchased the device: Custom Order
* Is this a custom build?: No
* Operating System: Windows 10
* Browser: Firefox

### Expected Behavior

Please describe the behaviour you are expecting.
When changing the pin assignment via the web config for an addon such as i2c, the previously used pins should become available for use.

### Current Behavior

What is the current behaviour?
When changing the pin assignment via the webconig for an addon, the new pins will be bound, but the old pins will also remain bound even though they are no longer in use.

### Steps to Reproduce

Note: Running these steps will permanently lock your pins out and can not be unlocked via normal means
1. Boot a board into the webconfig
2. Change the D+ pin for the usb passthrough.
3. Save and reboot into the webconfig
4. Go to the Pin Assignment page and notice how the previous and new pins both say they are assigned to an addon.

### Screenshots & Files

Please include any relevant screenshots or files here.  If you are having issues with a board please include images of the front and back of the board.

bsstephan:
Fixes #816

### DIFF
--- a/src/configs/webconfig.cpp
+++ b/src/configs/webconfig.cpp
@@ -106,32 +106,42 @@ static void __attribute__((noinline)) docToValue(T& value, const DynamicJsonDocu
 }
 
 // Don't inline this function, we do not want to consume stack space in the calling function
+static void __attribute__((noinline)) cleanAddonGpioMappings(Pin_t& addonPin, Pin_t oldAddonPin)
+{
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
That is much cleaner, thank you! Looks great

PR #894, opened 2024-03-16
Display update fixes

mikepparks:
* Fixed config pin defines for I2C blocks
* Fixed flip/mirror modes for display


arntsonl (approved):
Cool, easy fix

PR #898, opened 2024-03-17
Update turbo LED logic

FeralAI:
My previous Turbo update PR assumed the LED was active low, so the LED current would sink from +V. This caused incorrect and odd LED behavior for the more common use case of connecting LED+ to a GPIO pin.

This PR updates the default logic to use active high to drive the turbo LED, similar to #883, but also cleans up the LED state logic and allows overriding via board config using a define. This wasn't added as a togglable option via web config, but can be added if we think it's beneficial.

I also renamed the slightly ambiguous `turboButtonsPressed` variable to `turboButtonsMask`, which @SavageCore brought up when we discussed turbo behavior.


SavageCore:
Looks good. I can test tomorrow if needed

arntsonl (approved):
Cool, looks like an easy fix to the Turbo LED logic. Thanks @FeralAI and @SavageCore !!

PR #899, opened 2024-03-18
LayoutManager & button layout enhancements/fixes

mikepparks:
Fixes:
Addresses a bug with the Input History button glyph LUT that would cause the firmware to crash when run from a Debug build state.

Enhancements:
LayoutManager was created to allow sharing of built-in and custom layouts beyond the use of the OLED Display. This also includes two new API routes for the WebConfig:

* /api/getButtonLayoutDefs
  * Returns the list of ButtonLayout and ButtonLayoutRight definitions and their values.
* /api/getButtonLayouts
  * Returns the layouts for the current ButtonLayout and ButtonLayoutRight definitions, as well as the LED layout indexes.

This also includes enhancements to the button layouts to change from an unnamed array of parameters to named properties.  The defined button layouts will still use the previous structure for brevity, but the names exist to make the layout generation code easier to use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,7 @@ src/display/ui/screens/SplashScreen.cpp
 src/display/GPGFX.cpp
 src/display/GPGFX_UI.cpp
 src/drivermanager.cpp
+src/layoutmanager.cpp
 src/peripheralmanager.cpp
 src/storagemanager.cpp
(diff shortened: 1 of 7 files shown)


arntsonl (approved):
Nice, this is a good way of managing our layouts/input history, and keeping the debug from crashing. Those poor 264kb of RAM

PR #900, opened 2024-03-19
Analog sensitivity

tanakataroumaru1:
To improve the ease of use of the analog sticks, I added a sensitivity setting.
I have added a process to multiply the angle of the analog stick by a power factor to achieve the sensitivity shown in the image.
Please consider adding this process if you like.
(This is a machine translation.)

Analog_Sensitivity_Setting_a
Analog_Pow_Tilt


arntsonl:
Heads up, we are going to push this into the next release cycle as we are feature locking currently.

tanakataroumaru1:
Thank you for your reply.
I will rewrite the process to reflect the points you have pointed out.
(This is a machine translation.)

PR #904, opened 2024-03-22
Turbo state toggle updates

FeralAI:
This PR brings a few changes to Turbo activation and behavior. This attempts to make the turbo toggle behave consistently whether a user presses the turbo activation button before or after the action button to toggle. This also reverts the behavior of clearing gamepad outputs when the turbo toggle pin is active. This makes the behavior in line with some other commercial products with turbo, namely the 8bitdo products I've tested.

@TheTrainGoes please test when you get a chance and let me know how the new behavior feels.


arntsonl (approved):
Moar approval to moar turbo

PR #906, opened 2024-03-22
Allow some hotkeys to be held again

### LINKED ISSUE #876: Fn + Activator Button = Output Button, but cannot output long press
### Prerequisites

### Context

* Firmware Version: Compiled from latest [main](https://github.com/OpenStickCommunity/GP2040-CE/commit/53099a722f1ed21f222825b75e32c56c218239a7)
* Name of device: -
* Link to where you purchased the device: -
* Is this a custom build?: -
* Operating System: Windows 11
* Browser: Chrome

### Expected Behavior

When using Fn as a shift key, for example, Fn + A = RT, it should be able to output long-press RT based on hold long the Button A is pressed.

### Current Behavior

Output only last for a split second

bsstephan:
Some hotkeys are just chords of a sort for normal button outputs, and should be allowed to be held. Some refactoring accidentally undid that and had all hotkeys be checked against the previously run hotkey, such that they all only ran once until the hotkey was let go. This restores the old functionality and also guards some additional hotkeys that don't need to be held and should only run once because they are toggles and/or they save to flash.

Fixes #876

### DIFF
--- a/headers/gamepad.h
+++ b/headers/gamepad.h
@@ -89,7 +89,7 @@ public:
 	}
 
 	/**
-	 * @brief Remote hotkey bits from the state bitmask and provide pressed action.
+	 * @brief Remove hotkey bits from the state bitmask and provide pressed action.
 	 */
 	inline GamepadHotkey __attribute__((always_inline)) selectHotkey(const HotkeyEntry hotkey) {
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Nice easy fix, going in!

PR #909, opened 2024-03-23
DDI improvements and cleanups

### LINKED ISSUE #725: Discussion/POC: simplify Dual Directional Input by removing some/most combination modes
With 0.7.6 getting multiple pins outputting the same direction button, and #724 potentially allowing one pin to output multiple buttons, I think we should talk a bit about the Dual Directional Input addon. Specifically, I think at least one of the four combine modes is unnecessary now, and perhaps three of them (all but None) are on the chopping block.

### Combination Mode Mixed

> Combines both the Gamepad directions and Dual Direction Input directions and allows for all 3 SOCD modes.

This essentially has DDI work as 4 extra directional buttons that contribute to the core gamepad output. I think this is no different than mapping those buttons as more core directions, come 0.7.6, so I think this mode can be removed. Note that "Dual D-Pad Mode" has no relevance to this mode.

### Combination Mode Gamepad

> Gamepad always takes over when pressed, otherwise Gamepad and Dual act independently.

I have never fully understood this feature. The Dual "half" of the output is configured to be LS/DP/RS specifically, independent of the core gamepad's setting. Take the following situation:

* core Gamepad in DP mode
* DDI Mode in RS mode
* DDI combination mode "Gamepad"

If you...

1. press Gamepad Up, you get DP Up *and* RS Up (this is probably a bug, but I don't know since when. I'm on v0.7.6-RC1)
2. press DDI Up, you get RS Up (okay)
3. press Gamepad Up and DDI Up, you get DP Up and RS Up (this is probably the intentional behavior, unlike the first situation)
4. press DDI Down, you get RS Down (obv.)
5. press DDI Down and Gamepad Up, you get DP Up *and RS Up* (unsure about which of this is desired, but this is Gamepad taking priority over DDI, and forcing DDI's Down to be core's Up, as described by the Combination Mode)

I think this is *supposed* to work like Dual Directional mode below, but right now it is way buggy...

### Combination Mode Dual Directional

> Dual always takes over when pressed, otherwise Gamepad and Dual act independently.

I don't understand the use case for this feature either, but it works more like how I thought it would.

* core Gamepad in DP mode
* DDI combination mode "Dual Directional"
* (note that Dual D-Pad Mode again doesn't matter here)

If you...

1. press Gamepad Up, you get DP Up
2. press DDI Up, you get DP Up
3. press Gamepad Up and DDI Down, you get DP Down (order does not matter)
4. press Gamepad Left and DDI Left, you get DP Left (obv.)
5. press Gamepad Left and DDI Right, you get DP Right (order does not matter)

This is more clearly the DDI inputs taking precedence over the core gamepad inputs. However, I'm not sure what the use case for the behavior is. Again with multiple pin to button mappings possible now, I'm not sure how this precedence is any different than First Input Wins SOCD or Second Input Wins SOCD depending on your desired form of buttons.

### Combination Mode None

> Gamepad input and dual directional act independently of each other.

This does exactly what it says, and is maybe the only mode that *intentionally* uses the Dual D-Pad Mode setting. Core gamepad and DDI directions act independently, each getting SOCD cleaned independently (unless they're both contributing to the same output mode, in which case they get SOCD cleaned as one). This essentially allows for an extra set of directionals that *can* be mapped to something different than the core gamepad, distinguishing this mode (and maybe only this mode) from the ability to map multiple pins to the same direction in v0.7.6.

## Conclusion

My opinions below, but I intend to act on this, so discussion welcome before I go gutting code:

* Mixed Combination Mode: remove the mode, it's obsolete. Migrate mixed users to more non-DDI directional mappings.
* Dual Directional Combination Mode: remove the mode, it seems like a more complicated way to achieve something achievable in core, absent someone (pinging @arntsonl) explaining the use case for this better than I can understand it.
* Gamepad Combination Mode: remove the mode, as per Dual Directional, or else if it's important, fix it.
* None Combination Mode: keep it, this has a unique feature beyond what's possible in core, at least until we refactor things further.
* Dual D-Pad Modes: keep them all, they are relevant to None Combination Mode.

I think removing the three combination modes above will greatly simplify the addon, which is pretty hard to read, and further the ability for us to fold DDI into core in some fashion in the future. Comments desired.

### LINKED ISSUE #823: DDI in DDI mode does not remember previous state of cleaned button presses on gamepad
### Prerequisites

- [X] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [X] I have checked the documentation [HERE](https://gp2040-ce.info/) and found no answer
- [X] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [X] I am reporting the issue to the correct repository

### Context

* Firmware Version: v0.7.7 Haute42
* Name of device: Haute42 S16
* Link to where you purchased the device: X
* Is this a custom build?: No
* Operating System: X
* Browser: X

### Expected Behavior

* map down to DDI down
* set combination mode to DDI
* put on last win SOCD
* hold left then right, right takes SOCD priority
* while still holding left + right press down to override with DDI down

Expected - release DDI down, left + right still held, right trigged based on previous holding and SOCD cleaning

### Current Behavior

Actual - release DDI down, left + right still held, neither show up since firmware does not know which one was last.

bsstephan:
This...

1. Cleans up the code for the DDI addon, especially moving most of the logic formerly in `preprocess()` into `process()`, and subsequently making `process()` easier to read.
2. "Soft deprecates" Combination Mode Gamepad Override, which was at best a duplicate of DDI Override with different pin mappings, but also had some logic issue somehow, so it was also buggy (see discussion in #725)
3. Fixes #823 
4. Resolves #725


arntsonl:
I'm going to leave this one until after release as DDI is such a funky beast.

Would also be very worth asking our users who many folks used mixed in a stick/button combo and maybe if its used by Tekken folk.

arntsonl (approved):
Good changes, documents dual directional and gives us better control over the inputs.

PR #910, opened 2024-03-23
Analog sensitivity setting

tanakataroumaru1:
This is a suggestion regarding the addition of an analog stick sensitivity setting
I have re-written the code again based on your previous suggestions
I don't think it's appropriate to change the ADC value after the dead zone process, so I've done it in the current location.
However, the ADC values have changed significantly and the magnitude values need to be recalculated.
I am struggling with this point. If you have an evaluation, please let me know.
(This is a machine translation)


tanakataroumaru1:
I was not aware of the conflict and will correct it.

Issue #709, opened 2023-12-26
Wii Extension Nunchuck, has issues with left direction


Joomaton:
### Prerequisites

Please check the following before posting an issue / bug report.

- [x] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [x] I have checked the documentation [HERE](https://gp2040-ce.info/) and found no answer
- [x] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [x] I am reporting the issue to the correct repository

### Context

Please provide all relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.  

* Firmware Version: 0.7.6 Waveshare
* Name of device: Waveshare RP2040-Zero
* Is this a custom build?: yes

The nunchuck is a cheap AliX knockoff.
Note: The stick has a slightly higher resistive value on the x axis when pulled all the way left than compared to the y-axis

### Expected Behavior

Turning the analogue stick left should result in direction left.

### Current Behavior

The direction is registered as long as the analogue stick is not pushed all the way to the left.
Pushing it all the way to the left results in no direction. Also top-left and botton-left show up als top and bottom.

mikepparks:
Some adjustments were made in v0.7.8 to handle issues with the left directions and analog precision overall for Wii extensions.

PR #911, opened 2024-03-23
#709 - Fixed macros no longer working when more than one macro exists

ACavuoto:
Most of the line changes are code formatting. The main change is on line 167, so that the reset function is called also in a different scenario.

Tested on my Haute42 and seems to be working


arntsonl:
Unfortunately there is something deeper in macros causing an issue. I'll take a look and get it fixed up with our pin system.

PR #913, opened 2024-03-24
update reverse input to work on the dpad state rather than pin masks

### LINKED ISSUE #822: Enabling Reverse Input Button Breaks SOCD
### Prerequisites

### Context

* Firmware Version: GP2040-CE : v0.7.7
* Name of device: SallyBox, Flatbox R1, etc.
* Link to where you purchased the device: https://fightboxarcade.com/collections/fightbox-r-game-controllers / https://www.aliexpress.us/item/3256805359350735.html
* Is this a custom build?: No
* Operating System: Windows 11
* Browser: Firefox / Chrome

### Expected Behavior

When turning on the addon for Input Reverse, pressing the button should reverse the input as defined in the reverse input rules. Additionally, existing functionality should not be impacted when this button/mode is active.

### Current Behavior

When the addon for Input Reverse is enabled, all SOCD modes break. All buttons can be held down at any time without SOCD cleaning. This is irrespective of the Input Reverse button being pressed.

### Steps to Reproduce

1. Set SOCD cleaning mode to any mode other than "Off"
2. Reboot device and see that SOCD presses are properly handled
3. Enable a button for Input Reverse
4. Reboot device and see that SOCD presses are not properly handled

bsstephan:
using the core gamepad pin masks meant that anything that was adding to/modifying dpad state was being overwritten because they probably fired before reverse input, and thus their change to the gamepad state was being dropped because the GPIO pins themselves were inactive, so the reverse input logic thought the outputs were inactive and assigned to the dpad e.g. 0 | 0 | 0 | 0 | 0

this fixes reverse input + DDI entirely, and also allows the proper SOCD of the dpad(s) to be inverted rather than dropped.

note that reverse input only works against the dpad, I don't know if that's intended or not, but I'm leaving it as is.

Fixes #822

### DIFF
--- a/src/addons/reverse.cpp
+++ b/src/addons/reverse.cpp
@@ -62,13 +62,12 @@ void ReverseInput::process()
     update();
 
     Gamepad * gamepad = Storage::getInstance().GetGamepad();
-    Mask_t values = gamepad->debouncedGpio;
 
     gamepad->state.dpad = 0
-        | input(values & mapDpadUp->pinMask,    mapDpadUp->buttonMask,      mapDpadDown->buttonMask,    actionUp,       invertYAxis)
-        | input(values & mapDpadDown->pinMask,  mapDpadDown->buttonMask,    mapDpadUp->buttonMask,      actionDown,     invertYAxis)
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Totally agree on this one, good find.

PR #914, opened 2024-03-24
Display updates for layout changes

mikepparks:
* Fixes to displaying layout changes while in webconfig
* Added profile number banner display on profile change
* Updated some whitespace inconsistencies


arntsonl (approved):
Looks good! Interesting use of unique_ptr, will be good to look into for other pieces of our code

PR #915, opened 2024-03-24
#709 - Macro bug fix

ACavuoto:
Resets macroposition before returning


arntsonl:
We moved a different direction with Macros, it needed a rewrite pretty bad.

Replaced this with PR#922, no longer need one-line fixes for the old code.

PR #917, opened 2024-03-24
Migrate the turbo pin to profiles, not just the core mappings

### LINKED ISSUE #916: Turbo migration to GpioMappings needs to also migrate into profiles
Beta tester report against pre-0.7.8.

### Expected Behavior

* Turbo pin is assigned to the turbo GPIO action on all four profiles

### Current Behavior

* Turbo pin is only assigned to turbo in Profile 1, Profiles 2-4 still show it assigned to addon

I also think we should remove the comment `// Check for additional migrations for features 0.7.6+` and rename `checkAdditionalMigrations` as it gives the impression that all future migrations will go in that method, which I don't think is necessarily a good idea as future migrations may be complicated and/or non-repeatable, and eventually this method will get obnoxious to run at every boot. I would make it like the other migrations.

bsstephan:
Also standardizes the profile naming scheme a bit for clarity.

Fixes #916.

### DIFF
--- a/src/config_utils.cpp
+++ b/src/config_utils.cpp
@@ -1223,18 +1223,23 @@ void gpioMappingsMigrationProfiles(Config& config)
     config.migrations.buttonProfilesMigrated = true;
 }
 
-// Check for additional migrations for features 0.7.6+
-void checkAdditionalMigrations(Config& config) {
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
This looks good! Following this convention for macros

PR #919, opened 2024-03-24
recreate turboPinMask when doing reinit() (when profiles are changed)

### LINKED ISSUE #918: Turbo doesn't implement reinit(), won't work properly with profiles
### Expected Behavior

* Turbo pin changes in profiles are respected.

### Current Behavior

* If the turbo button pin changes between profiles, the change is not reflected in the addon, potentially causing odd behavior depending on what's mapped where in the new profile.

### Context

Just saw this one while looking at #916.

bsstephan:
Fixes #918

### DIFF
--- a/headers/addons/turbo.h
+++ b/headers/addons/turbo.h
@@ -89,14 +89,14 @@
 class TurboInput : public GPAddon {
 public:
     virtual bool available();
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
This is great, following same convention for macros

Issue #897, opened 2024-03-16
Cut off elements in Custom LED config page


goblinpope:
The color picker cuts off for inputs at or below the height of left/right (stick layout) or bottom row with no ability to scroll the page down or drag the picker back in full view.

![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/17583270/436c58f6-c6e4-4495-a212-13f94ff919b0)
![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/17583270/e4579419-767a-463a-9215-6af1bcf257cd)
![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/17583270/d3c1b399-3394-4cdd-a536-09a8772655bf)

bsstephan:
This has been confirmed to have been resolved by #920, closing. Thanks!

PR #920, opened 2024-03-26
Enables scroll for color picker when out of view

Pelsin:
This should also solve the same issue in led config page.
A quick fix for #897

https://github.com/OpenStickCommunity/GP2040-CE/assets/5345892/15742081-a597-4979-88e8-ec2f56895e2f


arntsonl (approved):
Nice, that's an easy win!

PR #922, opened 2024-03-27
Macro Rewrite

arntsonl:
This is a rewrite of our macro system.

This moves our macro pins over to the Pin Manager panel, including our Macro Button for holding Macro + Gamepad button inputs.

The Macro page has been rewritten to use tabs instead of a large list.

Macro CSS has been cleaned up to follow proper d-flex and formatting.

Lots of other changes, but functionality should be the same as before but with Pin Mapping involved.


arntsonl (commented):
Lots of changes for Macro including display updates and some logic additions

jack2game:
Hi, I'm not sure if this is intended, but it seems that
`return this->drawSticklessButtons();`
can be written as
`return drawSticklessButtons();`
so that the `getRightLayout` codes will be consistent with `getLeftLayout`



> Current Right:
> https://github.com/OpenStickCommunity/GP2040-CE/blob/dea263df3c059c73a4c14a7216e0692fc0a90159/src/layoutmanager.cpp#L121

> Current Left:
> https://github.com/OpenStickCommunity/GP2040-CE/blob/dea263df3c059c73a4c14a7216e0692fc0a90159/src/layoutmanager.cpp#L51

mikepparks:
I would propose that style and code consistency issues would be best left to the team to decide/act on. Good feedback but whether it gets changed or not is undecided.

PR #926, opened 2024-03-31
Change analyze command, dev dep. Update www readme

Pelsin:
Suggestion to use another bundle vizualiser without forcing devs to install it.

e.g
Screenshot 2024-03-31 at 06 51 01


arntsonl (approved):
Works fantastically on my machine, looks great!

PR #936, opened 2024-04-03
Add styling for disabled state

Pelsin:
Fix pin mapping page styling disabled state
Screenshot 2024-04-03 at 18 59 34
Screenshot 2024-04-03 at 18 59 41


arntsonl (approved):
Clean! Yeah that makes sense

PR #939, opened 2024-04-04
I2C peripheral device fixes

mikepparks:
Wii 
* Added tweaks for Wii analog calibration when calibration values do not exist.

* Fixed issue with Wii config values on strum up/down.
* Fixed issue with Wii guitar orange button being incorrectly mapped in source data.

Display
* Added display viewport object for resizing the non-status bar areas.
* Implemented button layout scaling when input history is enabled.


arntsonl (approved):
Very neat changes!

PR #940, opened 2024-04-05
Additional I2C peripheral tweaks

mikepparks:
Wii
* Les Paul Wiitar calibration now allows full use of whammy bar

Display
* Elements on button layout screen now center horizontally when scaled down.


arntsonl (approved):
Looks good!

Issue #927, opened 2024-03-31
Can't set color to function in custom theme page


Pelsin:
Adding this, so i don't forget. Or if someone else wants to check it out!

Tested on main haute config, latest version and in web dev mode.

### Current Behavior
Type error in web config when trying to set color to function in custom theme page.
Color picker does not show on press.

### Screenshots & Files
<img width="817" alt="Screenshot 2024-04-01 at 00 23 20" src="https://github.com/OpenStickCommunity/GP2040-CE/assets/5345892/9f547812-725f-4230-b95e-99eea632aa73">

<img width="532" alt="Screenshot 2024-04-01 at 00 23 14" src="https://github.com/OpenStickCommunity/GP2040-CE/assets/5345892/2f607aaf-5c6e-4f2b-ab15-b3f229b83f81">

PR #943, opened 2024-04-05
Custom led theme

Pelsin:
We have not implemented being able to set color to the function button. this results in an error when pressing it.
Unsure if it was ever intended, the rendering was a biproduct of having an incorrect const.
Api does not support it either, i've removed it so the user doesn't think it's possible to change it until we are done with LED rework.

Before
Screenshot 2024-04-06 at 00 10 49

After
Screenshot 2024-04-06 at 00 11 01

#927

### DIFF
--- a/www/src/Data/Buttons.js
+++ b/www/src/Data/Buttons.js
@@ -162,7 +162,7 @@ export const BUTTONS = {
 	},
 };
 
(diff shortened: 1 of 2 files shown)


arntsonl (approved):
Agreed, I think this was a leftover and we can add it back in

PR #946, opened 2024-04-07
Wii guitar range adjustments and analog trigger handling

mikepparks:
* Fixes issue where controller modes that support a single direction of analog movement (triggers mainly) would not function.
* Removed previous fallback calibration for guitars which was based on a broken whammy bar.


arntsonl (approved):
These are not something we need to put in the web config right? Looks good, I imagine the wii stuff has some magic numbers.

Issue #928, opened 2024-04-02
Keyboard keys are not remamppable after PR #890


octopuserectus:
### Expected Behavior

Remapping a key in Keyboard Mapping should take an effect
### Current Behavior

Remapping a key takes no effect after PR #890 

### Context

* Firmware version: v0.7.8-RC1
* Board config name: Haute 42
* Name of device: Haute Pad S16
* Link to where you purchased the device https://aliexpress.com/item/1005006421550898.html

### Steps to Reproduce

1. Boot into web-config
2. Go to /settings
3. Choose keyboard from the dropdown
4. Change any key to a non-default value
5. Save changes
6. Reboot into hid-kb mode
7. Press the key we just changed---it's unchanged

arntsonl:
Cool ty, we'll look into it. I'm assuming I missed something on the www react code when I ported it over to that page.

octopuserectus:
Closing since it's fixed with #947

PR #947, opened 2024-04-07
Keyboard settings issue

Pelsin:
This PR aims to fix fetching/validating/saving of key mappings on settings page.

One issue was that we had nested forms. therefor the key mapping form was never submitted.
As there can be only one!

As quite a bit was being changed in the file so i decided to format the file according to .editorconfig.
I kept the formatting to one [commit](https://github.com/OpenStickCommunity/GP2040-CE/commit/44ee9d3e833b045b3fa5b9132f9f53ddbbf5a0d2)

The important commits are:
https://github.com/OpenStickCommunity/GP2040-CE/commit/f4f0b883613d00ca296259c10b1e61f162c24495
https://github.com/OpenStickCommunity/GP2040-CE/commit/a327f79e9dba2440e70e605c650cfce1c406dde9
https://github.com/OpenStickCommunity/GP2040-CE/commit/7fe34263c185aeb739f82df71316c259d8b695f7

## Important
When saving input mode "keyboard" validation will occur before saving, so if the validation fails the input mode wont be saved.
Alternative change would be to first save input mode then validate keymappings.

 #928

### DIFF
--- a/www/src/Components/KeyboardMapper.jsx
+++ b/www/src/Components/KeyboardMapper.jsx
@@ -4,6 +4,7 @@ import FormSelect from './FormSelect';
 import { KEY_CODES } from '../Data/Keyboard';
 import { BUTTONS } from '../Data/Buttons';
 import boards from '../Data/Boards.json';
+import cloneDeep from 'lodash/cloneDeep';
 
 const KeyboardMapper = ({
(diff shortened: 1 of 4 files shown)


arntsonl (approved):
Looks good! Everything looks great!

PR #949, opened 2024-04-08
Fix Open_Core0 and Open_Core0 WASD configs

TheTrainGoes:
This PR updates the config files for the Open_Core0 and Open_Core0 WASD so they function as intended.

There is also a small update to buttonlayouts.h as S1 was represented for both S1 and S2 on the Open_Core0 WASD layout.


arntsonl (approved):
Yep, good fixes!

PR #956, opened 2024-04-10
Buttons Angled layout using wrong definition

mikepparks:
Fixed incorrect function for "Buttons Angled" right-side layout.


arntsonl (approved):
Easy thing to mix up, looks good!

PR #964, opened 2024-04-12
Update and supplement Simplified Chinese language files

huchang47:
Update and supplement Simplified Chinese language files


SavageCore:
Thanks/谢谢!

There are errors in `AddonsConfig.jsx` and `SettingsPage.jsx`, fix those and it looks good. Be sure you run `npm run dev` from the `www` folder, to check the language files compile.

huchang47:
Thanks for pointing out the error. I'm not a C++ programmer and I still have some problems with the compilation environment, I apologize for not catching the error in time. I will fix it and resubmit it in time.

PR #969, opened 2024-04-13
Invert GPIO Config

speedypotato:
This is nice because it supports normally closed switches and optical hardware.

Still needs Web Config which is not my specialty, might chip at it later.


arntsonl:
I'm going to move this to draft and see if we can come up with a good way to merge this into our gpio set.

bsstephan (changes requested):
I have left review comments (and now remember how GH works).

arntsonl:
I'm going to close this one for now and we can revisit when your board is ready and we are ready to do some inclusions of inversion.

Thank you for submitting this code! It's very interesting to see how all this worked.

PR #974, opened 2024-04-17
Display driver fixes for non-functioning displays

mikepparks:
Using this WIP to track and update issues with displays that do not function with the new TinySSD1306 library.  As it stands, affected displays use the SH1106 chipset, which uses similar initialization commands, but requires a different method of loading frame buffer data.


arntsonl (approved):
Looks fantastic! Should hopefully help with our SH1106 issues and add some good layouts for next release.

Issue #976, opened 2024-04-17
Macro - Hold Press + Input results in flickering input rather than continuously held input


InfraredAces:
### Expected Behavior

After configuring the macro with a single line and combination of inputs, while holding the macro trigger, it holds continuously holds the input down until the trigger is released.

### Current Behavior

After configuring the macro with a single line and combination of inputs, while holding the macro trigger, the input will be held, but will occasionally flicker off

### Context

Firmware Version: 0.7.8
Board Config Name: GP2040-CE_0.7.8_Haute42COSMOX.uf2 
Device Name: S16
Operating System: 6.8.4-arch1-1 (64-bit)
Browser: Chromium

### Steps to Reproduce

1. Configure Macro to a pin as seen in the screenshot below
2. Press and hold button for macro
3. View input flickering using https://hardwaretester.com/gamepad

### Screenshots & Files

![image](https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/bd7b22bd-5a0f-4ade-991c-b6a09fc68ce6)

https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/a288a196-7feb-48bf-a609-219378c6605f

FeralAI:
@InfraredAces can you test out a build from #977 and see if that resolves the issue?

InfraredAces:
@FeralAI Using the build from #977 does seem to resolve the issue. Help the Macro for 10+ seconds and I did not find any flickering through the gamepad tester

Issue #975, opened 2024-04-17
Unable to Save Changes After Deleting Line in Macro Configuration (0.7.8)


brentaro:
### Expected Behavior

Unable to save changes after deleting a line (double clicking explosion icon) in Macro Configuration. The save button does not work

### Current Behavior

Unable to save changes after deleting a line (double clicking explosion icon) in Macro Configuration. The save button does not work. The "Saved! Please reboot your device" message does not appear and it doesn't save any changes. Also, if deleting a line above other lines, the delete icon for subsequent lines does not work as well.

### Context

Haute S16 on 0.7.8 (GP2040-CE_0.7.8_Haute42COSMOX.uf2)

### Steps to Reproduce

1. Open Web Configuration
2. Navigate to Configuration > Macros Configuration
3. Create a Macro
4. Save
5. Navigate to another page and back to Macros Configuration.
6. Attempt to delete a line by double clicking the explosion icon.

The Save button should not display a message to reboot. And changes will not be saved.

### Screenshots & Files

N/A

TheTrainGoes:
This has been fixed via #977 .  No reply from original poster.

PR #977, opened 2024-04-17
Macro Restart and UI Fixes

FeralAI:
The macro logic was processing macro inputs into gamepad state prior to checking if the macro needed to be reset. This resulted in repeating macros missing applying gamepad state at the end of the macro. Swapping the check for resetting the macro to before the gamepad inputs are applied and creating a `restart()` method solves this issue.

Also fixed an issue on the macro config page when adding and deleting macro actions. The deleted lines weren't being filtered out, causing no console errors but preventing any changes to the macro actions array. Fixed up the React key errors in the console for this page as well.

Addresses https://github.com/OpenStickCommunity/GP2040-CE/issues/975 & https://github.com/OpenStickCommunity/GP2040-CE/issues/976

### DIFF
--- a/www/src/Pages/InputMacroAddonPage.tsx
+++ b/www/src/Pages/InputMacroAddonPage.tsx
@@ -98,13 +98,10 @@ const FormContext = () => {
 		}
 		fetchData();
 	}, [setValues]);
-/*
+
 	useEffect(() => {
-		async function setData() {
-			await setValues(filterMacroInputs(values));
-		}
-		setData();
-	}, [values, setValues]);*/
+		setValues(filterMacroInputs(values));
+	}, [values, setValues]);
 
 	return null;
 };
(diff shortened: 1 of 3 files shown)


www/src/Pages/InputMacroAddonPage.tsx
```diff
@@ -98,13 +98,10 @@ const FormContext = () => {
...
 	}, [setValues]);
-/*
+
 	useEffect(() => {
-		async function setData() {
-			await setValues(filterMacroInputs(values));
-		}
-		setData();
-	}, [values, setValues]);*/
+		setValues(filterMacroInputs(values));
```
arntsonl, line 103: this will crash your browser if you type in the name of the macro too fast.

I think this is a React bug, so we just need to implement some easy rate-limit:

`if ( time_since_update < rate_limit ) {
       updateValues = values;
       setTimeout(); wake-up
} else {
       setValues(filteredMacroInputs(updateValues));
}`
or something

FeralAI: I refactored the existing logic to pass the form update handler to the child component instead of trying to rely on useEffect in the child component. This allowed the removal of the EMPTY_INPUT logic on the page, and was replaced with a real delete (splice) of the input from the array.

arntsonl (approved):
Nice!

PR #984, opened 2024-04-21
Update lock file

Pelsin:
Saw another PR that had package.lock diff without any changes in package.json, wanted to make sure it's correct and separate it in it's own PR


arntsonl (approved):
yeah as much as I like to remove lock files from .github, this is good!

PR #987, opened 2024-04-27
Fix LED order reverting on save and cleanup

FeralAI:
This fixes an issue on the LED config page where after changing the order and saving, the Button Order list would revert to the initial page state. The save would still happen and the new order would show after a refresh, but was a bit confusing.

Also cleaned up a bit of the value conversions and centralized the data source creation for the draggable list.


arntsonl (approved):
Good fix! I've heard a few people tell me about this issue.

PR #988, opened 2024-04-27
Replace Axios with custom HTTP service to reduce bundle size

FeralAI:
This replaces the Axios NPM module with a custom Http service using `fetch` instead. We weren't doing anything crazy with the arguments, so it's a pretty simple implementation. Did some testing on my debug board and GET and POST behavior seemed to be working fine.

Here are the WebUI bundle sizes before and after:

```
BEFORE
vite v4.3.9 building for production...
✓ 1183 modules transformed.
build/index.html                     0.97 kB │ gzip:   0.55 kB
build/assets/index-96efa081.css    235.13 kB │ gzip:  32.79 kB
build/assets/index-f177e639.js   1,117.47 kB │ gzip: 342.96 kB

AFTER
vite v4.3.9 building for production...
✓ 1141 modules transformed.
build/index.html                     0.97 kB │ gzip:   0.55 kB
build/assets/index-96efa081.css    235.13 kB │ gzip:  32.79 kB
build/assets/index-54c80381.js   1,088.42 kB │ gzip: 331.65 kB
```

Saves 11.31 kB on the bundle.


Pelsin:
Somehow i thought it would be more than 11.31 kB :)
But super nice, love the class with the same api

FeralAI:
One minor update, I moved the `fetch` calls into the try/catch blocks so it wasn't unhandled on some unexpected failure.

FeralAI:
> Somehow i thought it would be more than 11.31 kB :)

It does save about 29 kB from the generated JS bundle, but I was counting after compression since it will be around 11 kB we save from in the firmware.

www/src/Services/Http.ts
```diff
@@ -0,0 +1,43 @@
...
+			});
+
+			const json = await response.json();
+			return Promise.resolve({ data: json });
+		}
+		catch (err) {
+			return Promise.reject(err);
+		}
+	}
+}
```
asturur, line 41: if it is bundle size you want to save, post and get are nearly identical you can expose them as the same function wrapped with an option parameter:

```ts
async post(url: string, body: unknown, headers: object = {}) {
  this.fetchWrapper(url, {
    headers,
    method: 'POST',
    body: JSON.stringify(body),
  });
}
```

where fetch wrapper is basically the body of post or get with some extra spread operator in the optios

arntsonl (approved):
Looks great! I like it, easy win and clean change over.

www/src/Services/Http.ts
asturur: yes this would have give you at most other 200 bytes less. at most. between 100 and 200 probably. but felt like repeated code, so that was nice too

FeralAI: A better argument for consolidating the `fetch` calls is that data transformation and error handling is the same for `get()` and `post()`. The next time I'm working on web updates I'll take a looking at optimizing this.

PR #989, opened 2024-04-27
Navbar cleanup and enable mobile menu

FeralAI:
This PR tightens up the navbar styling and spacing, and also enables the use of the collapsed nav for mobile when screen is less than 768px.

Summary of changes:

* Updated logo and made that the brand icon/area of the navbar, logo file slightly smaller now.
* Removed `Home` menu option and rolled that into the logo.
* Slightly reorganized the navbar actions (language, theme, etc.)
* Removed the inline styles on the navbar actions and used a parent container style for alignment
* Default the nav dropdowns to open toward the right to remain visible on smaller screens, since they're pinned to that side of the screen
* Removed the `Danger Zone` button. As much as I'm going to miss that remanant of the past, I think it caused more confusion than good. This has been rolled into the end of the `Configuration` menu
* Renamed a few button labels to reduce space taken: `Nintendo Switch` is now `Switch`, `DirectInput` is now `DInput`.

Here's a comparison of the extended menu view:

OLD:
image

NEW:
image

The old mobile view for the menu would make all the nav list items disappear:

image

This is fixed and the hamburger menu icon will now appear:

image

And here's a preview of the expanded mobile nav:

image

There are probably some minor things to address going forward, but those can be taken care when we address the responsiveness of the page content.


arntsonl (approved):
Nice clean changes and easy wins

PR #994, opened 2024-04-30
Analog rewrite

NickGuyver:
Fix bugs with two sticks only performing circularity on latter, and rapidly spinning two sticks would cause the former to jump. Rewrite to better protect against these kind of issues in the future. Other general cleanups and minor fixes.


arntsonl (approved):
Looks great!! I still need to check assembly and see if declaring a new struct type in C++ has to recreate the entire format but I don't see this causing any performance issues.

PR #996, opened 2024-05-03
Value of rotatory encoder jumping from 0.999 to -1

asturur:
Hi everyone :wave:

I have been trying the hw-40 rotatory encoder with the pico board and i noticed that the encoder goes from -1 to 1 but then jump abruptly to -1 again and doesn't move from there anymore if not by rotating backward ( is not a wrap around )

Removing the minus sign from the function call fixes the issue but **does swap** the rotation direction.

I do think the rotation swapped is more natural, clockwise on the right, counter clockwise on the left, it makes sense for example for a steering wheel because is what we are used to, but i m not going to argue which one i better i can understand is not an absolute point of view.

Encoder addon settings:
image

## current 0.7.8 release:

Testing on https://hardwaretester.com/gamepad turning the knob 180+ degree counter clockwise and then going back to center and adding 180 clockwise

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/b45c2169-c9d0-48be-b563-05b7bc2d3306

## current patch in this branch, same movement

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/8c656740-ca8c-4076-b15f-8064c3ce3a1c

 compiled code for testing [GP2040-CE_0.7.8_Pico.uf2.zip](https://github.com/OpenStickCommunity/GP2040-CE/files/15197144/GP2040-CE_0.7.8_Pico.uf2.zip)

This is my encoder:
image

And those are the specs:
https://www.elecom.sk/rotary-encoder-module-with-demo-code-for-uno-r3/

### DIFF
--- a/src/addons/rotaryencoder.cpp
+++ b/src/addons/rotaryencoder.cpp
@@ -103,13 +103,13 @@ void RotaryEncoderInput::process()
             uint32_t lastChange = now - encoderState[i].changeTime;
 
             if (encoderMap[i].mode == ENCODER_MODE_LEFT_ANALOG_X) {
-                gamepad->state.lx = -mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
+                gamepad->state.lx = mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
             } else if (encoderMap[i].mode == ENCODER_MODE_LEFT_ANALOG_Y) {
-                gamepad->state.ly = -mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
+                gamepad->state.ly = mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
             } else if (encoderMap[i].mode == ENCODER_MODE_RIGHT_ANALOG_X) {
-                gamepad->state.rx = -mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
+                gamepad->state.rx = mapEncoderValueStick(i, encoderValues[i], encoderMap[i].pulsesPerRevolution);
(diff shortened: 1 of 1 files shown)


asturur:
@mikepparks you committed the original code and of course it could also just be a matter of 2 different encoders behaving wildly differently. Do you remember the reason for the minus sign on the analog sticks but not on the triggers/dpad?

asturur:
[GP2040-CE_0.7.8_Pico-upto1.uf2.zip](https://github.com/OpenStickCommunity/GP2040-CE/files/15197355/GP2040-CE_0.7.8_Pico-upto1.uf2.zip)

Remove the -1 from the upper bound that at this point wasn't making sense anymore

arntsonl:
I'll wait for @mikepparks but if you want to link the uf2, a really easy way to do it on your branch is to go :  Actions > Cmake > Run Workflow and select your up to date branch. That will generate the uf2 list we use for releases.

image

asturur:
> I'll wait for @mikepparks but if you want to link the uf2, a really easy way to do it on your branch is to go : Actions > Cmake > Run Workflow and select your up to date branch. That will generate the uf2 list we use for releases.
> 
> image

Then i would have to upload them all? or just link the action in some way?

asturur:
Just to be clear, i m puzzled by this fix.
I pointing the finger at uint/int conversion and putting a negative number into a uint but my expertise in C is not so strong to say that is exactly the case.

It is definitely a fix for my use case

mikepparks:
I'm going to close this as I am currently waiting for these sorts of encoders so I can test the results myself. I don't exactly feel that changing the signing of the results is the fix, but is one avenue that I'll test. Please make an issue explaining the findings you outlined here so I can reproduce.

asturur:
yes i ll convert this to an issue.
Meanwhile, which are the encoders you used to develop the feature?

asturur:
issue: https://github.com/OpenStickCommunity/GP2040-CE/issues/998

PR #997, opened 2024-05-03
Add config for the ARC Accessibility Controller

TheTrainGoes:
This PR adds a config for the ARC Accessibility Controller


arntsonl (approved):
So cool!! This is huge for accessibility controllers

PR #999, opened 2024-05-04
feat(): Add a mode for the rotatory encoder to trigger 2 buttons

asturur:
This is the feature i originally wanted and i though i could now have it since most of the code is there.

For some arcade situations in which the DPAD is used and the analog controllers are not supported, having a rotatory encoder mapping on buttons is useful.

This is what i was thinking:
- i add a lastValue var where i keep the last state of the encoder, positive or negative.
- i add a new enum for the mode and relative web code to configure it
- i assign last value from the rotatory decode code and i keep it of the same value until it change or until a timeout pass ( using the available reset timer, otherwise the button press are like 1 ms long and most of the time are not captured, at least this is my understanding )

Using this configuration

image

I tested it only with https://hardwaretester.com/gamepad for now

## In action

Here is a test with an arcade game from mister that support either a spinner or buttons, in button mode:

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/e6965896-3c58-48f3-9e3a-545377d1b21c

## Why

The difference with the current modes is the absence of a central dead zone and the ability to go right and left immediately.

Both DPAD and Analog stick requires you to cross the center zone or to go through the reset after X millisecond.
If the reset is too quick you are always fighting with the center dead zone, if the the reset is too slow or the movement too quick you are not going back to the center first and you need to change direction by rotating a lot.

This different method is very useful for games that uses 2 buttons for rotation.

## How to use it

You select the new mode from the dropdown and you have to set the 'reset after' time.
The code will detect the change only for 1 tick ( around 1ms ) if you don't set a reset after time.
In this case the reset after time is cancelling the button press. A rotatory tick will push the button down and the reset time will push the button back up, the value you want to set depend on the game, some games will support continue press, others maybe not, and the more the reset after time is large the more the button stay pressed.

This is the first implementation i managed to make work and worked well, the rest can be built with feedback and other opinions

### DIFF
--- a/www/src/Locales/en/Addons/Rotary.jsx
+++ b/www/src/Locales/en/Addons/Rotary.jsx
@@ -16,6 +16,7 @@ export default {
     'encoder-mode-right-analog-y': 'Right Analog Y',
     'encoder-mode-left-trigger': 'Left Trigger',
     'encoder-mode-right-trigger': 'Right Trigger',
+    'encoder-mode-buttons': 'Buttons + / -',
     'encoder-mode-dpad-x': 'D-Pad Left/Right',
     'encoder-mode-dpad-y': 'D-Pad Up/Down',
 };
--- a/headers/addons/rotaryencoder.h
+++ b/headers/addons/rotaryencoder.h
@@ -116,6 +116,7 @@ private:
     EncoderPinState encoderState[MAX_ENCODERS];
     int32_t encoderValues[MAX_ENCODERS];
     int32_t prevValues[MAX_ENCODERS];
+    int8_t lastValue = 0;
     EncoderPinMap encoderMap[MAX_ENCODERS] = {
         {false, -1, -1, 24, ENCODER_MODE_NONE, -1, -1},
         {false, -1, -1, 24, ENCODER_MODE_NONE, -1, -1},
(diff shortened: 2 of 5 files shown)


www/src/Locales/en/Addons/Rotary.jsx
```diff
@@ -16,6 +16,7 @@ export default {
     'encoder-mode-right-analog-y': 'Right Analog Y',
     'encoder-mode-left-trigger': 'Left Trigger',
     'encoder-mode-right-trigger': 'Right Trigger',
+    'encoder-mode-buttons': 'Buttons + / -',
```
asturur, line 19: the label is just here for discussion, i didn't really think much of the name

mikepparks:
At the moment, the "Buttons +/-" option defines hard-coded button mask values, and I would like to avoid going that direction as it limits what can be used for the purpose.

asturur:
Yes i understand that. I didn't want in the weeds of defining a UI with custom buttons values before talking of the possibility of the function. So if we like the idea of having an extra mode i m more than happy to work on button customization values. I m not sure is really necessary but i can understand there could be a case in which you create this custom controller and l2/l1 in a game are already used by something else.

I can do that if we like the idea of this extra mode

mikepparks:
We'll revisit this once we make some improvements to how we handle GPIO actions.

PR #1001, opened 2024-05-05
fix() signal from axios.get is a specific AbortSignal

asturur:
I noticed by chance in the [http vs axios](https://github.com/OpenStickCommunity/GP2040-CE/pull/988) PR that the signal was axios.get was merged into an header, that seems a bug to me.

Here:
https://github.com/OpenStickCommunity/GP2040-CE/pull/988/files#diff-c906511881a407b1d4a0978d708b2841f02b1a72e52c5efd1210fb16628fce6eR637

the pin mapping page is indeed pushing up this wrong header:
image

This is possible in the 0.78 release:

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/0f7a7a20-b270-49f2-9757-4ce7bba27a62

But those buttons ( skip pin, cancel and the close icon ) are not functional in current master

### DIFF
--- a/www/src/Services/Http.ts
+++ b/www/src/Services/Http.ts
@@ -1,8 +1,13 @@
 // This is modeled after Axios to make an easier transition, so this may need more buildout
 // if more than simple GET and POST requests are required.
 
+type GetOptions = {
+	headers?: Record<string, string>,
+	signal?: AbortSignal
+}
+
 class Http {
-	async get(url: string, headers: object = {}) {
+	async get(url: string, { headers = {}, signal }: GetOptions = {}) {
 		try {
 			const response = await fetch(url, {
 				method: 'GET',
@@ -10,6 +15,7 @@ class Http {
 					'Content-Type': 'application/json',
 					...headers,
 				},
+				signal
 			});
 
 			const json = await response.json();
--- a/www/src/Services/Http.ts
+++ b/www/src/Services/Http.ts
@@ -22,9 +22,6 @@ class Http {
 			return Promise.resolve({ data: json });
 		}
 		catch (err) {
-			if (signal?.aborted) {
-				throw new Error('ERR_CANCELED');
-			}
 			return Promise.reject(err);
 		}
 	}
--- a/www/src/Services/Http.ts
+++ b/www/src/Services/Http.ts
@@ -22,6 +22,9 @@ class Http {
 			return Promise.resolve({ data: json });
 		}
 		catch (err) {
+			if (signal?.aborted) {
+				throw new Error('ERR_CANCELED');
+			}
 			return Promise.reject(err);
 		}
 	}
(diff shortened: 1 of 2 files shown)


asturur:
This fix doens't restore the functionality yet, but aborts the connection correctly.

asturur:
Ok it needed some extra line of handling to mimic axios response in this case.
It could be done also without but the error in case of abort with no handling is pretty much a long sentence and it seems that what MDN says doesn't match with what chrome says and so let's avoid it

www/src/Services/WebApi.js
```diff
@@ -639,7 +639,7 @@ async function getHeldPins(abortSignal) {
 		});
 		return response.data;
 	} catch (error) {
-		if (error?.code === 'ERR_CANCELED') return { canceled: true };
+		if (error.message === 'ERR_CANCELED') return { canceled: true };
```
asturur, line 642: this constant is used only here so i assume it was coming from axios

Pelsin (commented):
Nice find with the abort signal!

Added some comments for improvements

www/src/Services/WebApi.js
Pelsin: We should probably use fetch error 'AbortError' instead of creating our own/mimic axios

www/src/Services/Http.ts
```diff
@@ -1,21 +1,30 @@
...
 				},
+				signal
 			});
 
 			const json = await response.json();
 			return Promise.resolve({ data: json });
 		}
 		catch (err) {
+			if (signal?.aborted) {
+				throw new Error('ERR_CANCELED');
```
Pelsin, line 26: I don't think this is correct, fetch should already throw if aborted.

https://dom.spec.whatwg.org/#interface-abortcontroller

asturur: I don't think is a matter of correctness, yes is throwing and indeed we are in the catch block, the special case here was done to stick to 'ERR_CANCELED' that is a sort of custom identifier to identify the user abort operation. There are alternatives to do so, this was the less risky.

As i mentioned above the error returned changes between firefox and chrome. They are bot DOMException but different message

image

So if we don't like throwing our own error we can either:
- 1 use `err instanceof DOMException` as an identifier
- 2 use err.message containing the word 'abort' generic, hoping that is common to every browser as an identifier
- 3 do not use any identifier and call `abortGetHeldPins` after any error of `getHeldPins` but i m not sure of the implications
- 4 instead of calling 'abort()' we could call abort('-our-custom-abort-message') and that would make the fetch call reject a string instead of an error and we can just match the string ( in that case please pick a string ).


As a developer the current option and the 4th alternative option seems the best to me, the DOMException option is fine too with some risk of overlap from other errors that may happen.

Definetely your choice

asturur: Oh i see now your edit of the name, I didn't think of using the instance class name. that is probably the simples

Pelsin: When identifying DOMException i thought one used the property "name", is there a discrepancy between browsers there also?

arntsonl (approved):
Looks good

PR #1002, opened 2024-05-05
Update TinyUSB to latest

arntsonl:
This updates our TinyUSB to the latest library. This removes the get_report hack I added in while tinyusb was updated, and brings us up to 4/30/2024 in the TinyUSB library.


arntsonl:
I'm still having issues on this one for Xbox One

PR #1004, opened 2024-05-06
Rotary encoder joystick mode overflows incorrectly

### LINKED ISSUE #998: Issue with rotatory encoders HW-040
Hi Everyone, i recently got the 0.78 update for a rotatory encoder and i noticed it does not work as expected with the encoder i had home, the positive extreme of the rotation flips over to the opposite extreme.

### Expected Behavior

I can rotate the encoder and get a mapping on one axis like this

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/8c656740-ca8c-4076-b15f-8064c3ce3a1c

### Current Behavior

The encoder reset to -1 after reaching +1

https://github.com/OpenStickCommunity/GP2040-CE/assets/1194048/b45c2169-c9d0-48be-b563-05b7bc2d3306

### Context

Rotatory encoder used: hw-40
image

And those are the specs:
https://www.elecom.sk/rotary-encoder-module-with-demo-code-for-uno-r3/

Encoder addon settings in 0.78
image

### Steps to Reproduce

Get a rotatry encoder of the same kind
Connect it and configure it as by example or in similar way, try to rotate it on both directions, one direction is flawed

mikepparks:
Fixed issue where rotary encoder would occasionally max out in the opposite direction in analog joystick modes.
Fixes #998 .

### DIFF
--- a/headers/addons/rotaryencoder.h
+++ b/headers/addons/rotaryencoder.h
@@ -127,6 +127,8 @@ private:
     uint16_t mapEncoderValueTrigger(int8_t index, int32_t encoderValue, uint16_t ppr);
     int8_t mapEncoderValueDPad(int8_t index, int32_t encoderValue, uint16_t ppr);
 
+    int8_t getEncoderIndexByPin(uint8_t pin);
+    
     bool dpadUp = false;
     bool dpadDown = false;
     bool dpadLeft = false;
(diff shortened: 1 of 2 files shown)


asturur:
I tested the build works great, the bug is fixed and the rotatory control is usable now. By shaving off the 0 that does not work well with the complement of two that is used in the rotatory -> analog-stick mapping the jump is gone.

I also tested it with the original bound function by recompiling the branch and it behaves as good, so i wonder if it is convenient to dump the usage of the existing  function for the new bound check.

arntsonl (approved):
Looks good!

PR #1005, opened 2024-05-06
Update SGF configs to simplify

sgfdevices:
SGF is now using the same config for all controllers -- Bridget MX, Type F, PE, and Faust. The PCBs use different hardware, but the same config works for all of them.
We've removed the Bridget and Faust config folders and moved to the SGF Devices folder. This firmware has been tested on Bridget MX, Type F, PE, and Faust and passes tests on all of them.


TheTrainGoes (approved):
Looks good - Tested on both of my devices.

PR #1006, opened 2024-05-06
fix analog range

NickGuyver:
Fix for analog range to be complete at the loss of circularity, and some other minor fixes.


proto/config.proto
```diff
@@ -339,14 +339,15 @@ message AnalogOptions
...
     optional bool forced_circularity = 4;
-    optional uint32 analog_deadzone = 5;
-    optional int32 analogAdc2PinX = 6;
-    optional int32 analogAdc2PinY = 7;
-    optional DpadMode analogAdc1Mode = 8;
-    optional DpadMode analogAdc2Mode = 9;
-    optional InvertMode analogAdc1Invert = 10;
-    optional InvertMode analogAdc2Invert = 11;
-    optional bool auto_calibrate = 12;
+    optional uint32 inner_deadzone = 5;
```
arntsonl, line 342: This is a limitation of Protobuf, we can't change the enum of already defined variables. Otherwise it breaks.

So what we'll have to do is change analog_deadzone = 5 to inner_deadzone = 5 and outer_deadzone = 13;

arntsonl (changes requested):

arntsonl (approved):
Looks great!

PR #1007, opened 2024-05-07
Adding lodash types for IDEs that support TS intellisense

Pelsin:
Added as a dev dependency


arntsonl (approved):
Nice easy fix, thank you!

Issue #1008, opened 2024-05-12
Keyboard Host - Pressing combinations of Left Alt, Left Ctrl, and Left Shift result in no inputs


InfraredAces:
### Expected Behavior

Using the default Keyboard Host keycode-input mappings, pressing any combination of Left Alt, Left Ctrl, and Left Shift result in B1, B3, and B4 inputs

### Current Behavior

Using the default Keyboard Host keycode-input mappings, pressing any combination of Left Alt, Left Ctrl, and Left Shift result in no inputs at all.

### Context

Firmware Version: 0.7.7
Board Config Name: GP2040-CE_0.7.7_KeyboardConverter.uf2 
Device Name: Keyboard Converter
Purchase Link: https://github.com/OpenStickCommunity/Hardware/tree/main/Boards/Supported%20Boards/GP2040%20Keyboard%20Converter/Waveshare%20Zero-PCB

### Steps to Reproduce

Because Keyboard Host add-on does not work at all with 0.7.8, this was done with 0.7.7 and I was unable to confirm whether the issue persists with 0.7.8 firmware

1. Flash GP2040-CE controller that has a USB Host port with firmware
2. Connect keyboard
3. Press Left Alt, Left Ctrl, or Left Shift
4. Press Left Alt, Left Ctrl, or Left Shift, whichever key you didn't pressed in step 3

Changing the keycode-input mapping for inputs to other keycodes will cause it to work properly and, if other inputs are mapped to these keycodes, the issue will occur again.

### Screenshots & Files

Press and Hold Left Alt, then press Left Shift. Followed pressing Left Ctrl

https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/fba62169-104b-4b11-8dfe-6c002ca413b6

https://github.com/OpenStickCommunity/GP2040-CE/assets/44279698/afd66f81-f322-4d6f-aa36-ec5cc0b779d5

PR #1010, opened 2024-05-13
Keyboard Host fixes

mikepparks:
Added ``setup()`` call for Keyboard Host listener on Keyboard Host ``setup()``


arntsonl (approved):
Wow, that was a really dumb mistake I made on that one. Thank you for fixing that and adding the fix for #1008!

PR #1011, opened 2024-05-15
Update for Mavercade configs

TheTrainGoes:
This PR updates the Mavercade configs to split out into the older Rev1 config and newer Rev2 config.


arntsonl (approved):
Nice! gets us up to the latest with Mavercade

PR #1015, opened 2024-05-17
Button layout rectangle rotation

mikepparks:
Added rotation to rectangles using the angleStart property.
Renamed GP_SHAPE_DIAMOND to GP_SHAPE_LINE now that rectangles can rotate.


arntsonl (approved):
Nice!! This simplifies the diamonds quite a bit

PR #1017, opened 2024-05-18
Input history respects SOCD cleaning

mikepparks:
Input history now uses the processed gamepad for inputs to be in line with button elements


arntsonl (approved):
Looks good to me, makes sense with how our users expect to see outputs

PR #1022, opened 2024-05-23
Update Keyboard Converter Config

InfraredAces:


configs/KeyboardConverter/BoardConfig.h
```diff
@@ -58,7 +59,15 @@
 #define KEY_BUTTON_A2   HID_KEY_F2            // A2     | ~      | Capture | ~        | 14     | ~      |
 #define KEY_BUTTON_FN   -1                    // Hotkey Function                                        |
 
+#define USB_PERIPHERAL_ENABLED 1
+#define USB_PERIPHERAL_PIN_DPLUS 28
+#define USB_PERIPHERAL_PIN_ORDER 0
+#define USB_PERIPHERAL_PIN_5V -1
+
+#define HAS_I2C_DISPLAY 0
```
TheTrainGoes, line 67: Is this needed?  Have you tried compiling without 68 + 69?

TheTrainGoes (changes requested):

configs/KeyboardConverter/BoardConfig.h
InfraredAces: Yes, when compiling and flashing, I2C0 was enabled, but had no pins set and Keyboard Host would not function until I2C0 was disabled. By explicitly disabling both, there is no need for an intermediary step of going to web config to disable them.

TheTrainGoes: Odd

OK, in it goes.

TheTrainGoes (approved):

PR #1023, opened 2024-05-23
Ignore Docker and devcontainer files

InfraredAces:


arntsonl (approved):
That's interesting we have folks pulling in code in docker, but I'm okay with this

PR #1024, opened 2024-05-25
add boardconfig for ergoSHIFT Rev.2

mizma:
From previous PR:

> My general feedback from looking at your repo:
> 1 - There are parts of the repo listed as design not verified. This has traditionally lead to issues as people may make boards that could have issues. For this we would want to wait until the design is solidified, tested and ready for anyone to pickup and make.
> 2 - It requires the use of JFA SBM parts (keycaps and aux keys). Ideally there would be a 3D printed alternative to this. SBM parts can be quite expensive to get in some places.
> 3 - General concerns around this line You may need to adjust the rotation, placement of some parts in CPL, check before you order. It should look like the following.. If the CPL files are not lining up for services like JLC it will need to be fixed. This will lead to people not getting working boards or boards that could be missing parts.

1 - all parts have been verified and has been built by at least two other people.
2 - I have added 3D printed caps and have confirmed with third party print services (MJF) and confirmed they fit correctly with the design and choc v1 buttons
3 - CPL has been confirmed to work on my own build + two others without issue.  I left some notes to make sure to check before ordering due to concerns with JLC changing parts footprint data.


mizma:
board design repository: https://github.com/mizma/ergoSHIFT

arntsonl (approved):
Looks great!

PR #1025, opened 2024-05-28
PCF8575 I2C IO expansion module

mikepparks:
This implements the PCF8575 IO expansion module, a 16 pin digital IO expander. In its current form, the addon's configuration can define each pin as an input or output, and will handle the assigned Gamepad buttons accordingly.

Also included is a new I2C device driver class for eventual common functionality.  Currently, this class exposes a method for drivers to define their range of bus addresses, and provides a means to scan for the device on the I2C bus. In this implementation, it uses this ability to auto-detect a connected PCF8575 in its range of allowed addresses, and auto-configures itself if one is found.


arntsonl (commented):
I'll ask you over in Discord about those 2 things I saw but this is looking really clean!!!

PR #1028, opened 2024-05-28
Fix for B3 Button being unselectable in the hotkey setting page

MechLizard:
Both B2 and B3 buttons were assigned to value 24 on the hotkey web page drop-down. Resulting in B3 not being selectable.


arntsonl (approved):
Good find!

PR #1031, opened 2024-05-30
Add a hotkey to cycle between profiles.

### LINKED ISSUE #1016: Request: add hotkey function to cycle through profiles
**Problem:**
Currently, the only way to change button profiles is assigning them each a separate hotkey. Functionally, wanting to swap to one other profile means having to assign at minimum two different hotkeys. Meanwhile LED settings have 5 levels of brightness, 5 different animations, each with their own numerous variations, all of which are fully cyclable by hotkey alone.

**Solution:**
Add hotkey functions that cycle up and down through profiles. Also possibly make it to skip over profiles in the case they're identical to the first one. 

Alternatively, have profiles be assignable to input modes (e.g. set Xinput to use #1 by default, Switch uses #2, etc.). 

**Do you plan to work on this feature yourself?**
No.

MechLizard:
Added a hotkey option called "Next Profile". When assigned and pressed cycles to the next profile so that a button doesn't need to be assigned to each profile. 

For example, if on Profile 1 it changes to Profile 2, then to 3, then to 4, then back to 1.

I removed _lastAction = HOTKEY_NONE;_ from _gamepad.cpp_ since it is already assigned on initialization in _gamepad.h_ and seems to only cause hotkeys that use _userRequestedReinit = true;_ to trigger rapidly while the button is held down. 

For German and Portuguese translations, I had some friends that speak those languages confirm that they are correct translations. However, I used Chat GPT to get the simplified Chinese Translation. That should be considered a placeholder until someone that knows the language can take a look at it. 

Resolves #1016


mikepparks:
Going to ask that we split this into two PRs to separate contexts. Locale changes go to a separate PR from functional ones. Also, no AI translations. Thanks.

src/gamepad.cpp
```diff
@@ -81,7 +81,6 @@ void Gamepad::setup()
 		}
 	}
 
-	lastAction = HOTKEY_NONE;
```
bsstephan, line 84: Why this removal? It is necessary to keep hotkeys that would cycle or toggle through something if held (such as this one, incidentally) from doing something unpredictable. I don't think this should be removed.

headers/storagemanager.h
```diff
@@ -69,6 +69,7 @@ class Storage {
 	uint8_t * GetFeatureData();
 
 	void setProfile(const uint32_t);		// profile support for multiple mappings
+    void nextProfile();
```
bsstephan, line 72: Please match the tab/space style of the source file, thank you.

www/src/Pages/SettingsPage.jsx
```diff
@@ -244,22 +244,23 @@ const HOTKEY_ACTIONS = [
...
+	{ labelKey: 'hotkey-actions.b3-button', value: 26 },
+	{ labelKey: 'hotkey-actions.b4-button', value: 27 },
+	{ labelKey: 'hotkey-actions.l1-button', value: 28 },
+	{ labelKey: 'hotkey-actions.r1-button', value: 29 },
+	{ labelKey: 'hotkey-actions.l2-button', value: 30 },
+	{ labelKey: 'hotkey-actions.r2-button', value: 31 },
+	{ labelKey: 'hotkey-actions.s1-button', value: 32 },
+	{ labelKey: 'hotkey-actions.s2-button', value: 33 },
+	{ labelKey: 'hotkey-actions.a1-button', value: 34 },
+	{ labelKey: 'hotkey-actions.a2-button', value: 35 },
```
bsstephan, line 263: As earlier, you shouldn't change the existing enum values, so most of this diff should go away.

bsstephan (changes requested):
I have left review comments, in addition to echoing BRP's.

src/gamepad.cpp
MechLizard: That line would apply HOTKEY_NONE to lastAction immediately after the profile was loaded, causing it to rapidly apply the button's function dozens of times every second while the hotkey button is held down. This appeared to also happen with other hotkeys that used it, such as switching to a static profile, although we wouldn't notice it since applying profile 4 ten times still gives you profile 4. After I removed it it allowed me to press the button once to get it to apply only once. 

Although I think you are still correct, while it did seem to be working well for me I'm going to revisit the code to see if there is some other function that is setting it to HOTKEY_NONE and causing it to work correctly accidentally.

www/src/Pages/SettingsPage.jsx
MechLizard: Understood. I will change the enum values back and fix the spacing. Thank you for the feedback.

src/gamepad.cpp
bsstephan: If you find the first paragraph to be true when it's removed, then I think a refactor changed the behavior somehow. You may be right, so try more and I can try to take a look at main some time today too.

Either way though, indeed the behavior of this hotkey and other hotkeys like toggle 4-way should be that you hold it for any length of time and trigger it once, so focus on that and if you confirm the line should go, then that's a good find.

MechLizard: I looked at the code again and it appears that this module is called every loop of the main firmware loop. It overwrites the lastAction variable as long as there isn't a hotkey being pressed. So I feel that the line I removed is redundant. https://github.com/OpenStickCommunity/GP2040-CE/blob/affc3c498a37b558967bff91df2209103606df30/src/gamepad.cpp#L239-L266

The only issue I've noticed is that there is about a 1/4th of a second of delay before I can press the profile button again. But I don't see another spot that I could edit it to make it faster.

MechLizard:
I have removed the localization to add to another pull request if this one is approved as requested by [mikepparks](https://github.com/mikepparks) and  [bsstephan](https://github.com/bsstephan). 

I have also fixed the spacing and enum issues that [mikepparks](https://github.com/mikepparks) brought up. 

I believe that resolves everything mentioned. But I'd be happy to take another look if there is anything else out of line.

src/gamepad.cpp
bsstephan: 250 ms doesn't make much sense to me, but maybe the reinitialization blocks things for a bit; regardless I agree with your assessment otherwise.

proto/enums.proto
```diff
@@ -258,22 +258,23 @@ enum GamepadHotkey
...
-    HOTKEY_B4_BUTTON             = 26;
-    HOTKEY_L1_BUTTON             = 27;
-    HOTKEY_R1_BUTTON             = 28;
-    HOTKEY_L2_BUTTON             = 29;
-    HOTKEY_R2_BUTTON             = 30;
-    HOTKEY_S1_BUTTON             = 31;
-    HOTKEY_S2_BUTTON             = 32;
-    HOTKEY_A1_BUTTON             = 33;
-    HOTKEY_A2_BUTTON             = 34;
+    HOTKEY_NEXT_PROFILE          = 19;
```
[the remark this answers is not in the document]
bsstephan: Thanks for fixing this.

headers/storagemanager.h
bsstephan: Thanks for fixing this.

bsstephan (approved):

bsstephan:
My review is satisfied, looks good to me.

arntsonl (approved):
Looks good!

PR #1032, opened 2024-06-02
Fix indents in src/gp2040.cpp

sgysh:
I fixed indents in src/gp2040.cpp with editorconfig support.

### DIFF
--- a/src/gp2040.cpp
+++ b/src/gp2040.cpp
@@ -68,7 +68,7 @@ void GP2040::setup() {
 
 	// Setup Gamepad
 	gamepad->setup();
-	
+
 	// now we can load the latest configured profile, which will map the
 	// new set of GPIOs to use...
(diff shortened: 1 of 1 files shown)


arntsonl:
Let's please not do these types of one-off single file indent commits.

If you end up adding or modifying code in gp2040.cpp, please update with the correct indentation. I know we don't all follow the same indents, but I don't want to encourage single file PRs with 15-20 lines of indents.

If we decide to do a full lint of our C++ code later, we can commit it as a single large PR.

sgysh:
I understand.
The following part is especially difficult to read and I hope it will be fixed someday. @arntsonl 
https://github.com/OpenStickCommunity/GP2040-CE/blob/affc3c498a37b558967bff91df2209103606df30/src/gp2040.cpp#L372-L409

PR #1035, opened 2024-06-04
Adding Chinese localization for web config

rebornsick:
Supplement and improve Chinese localization.


SavageCore:
Thanks! I've marked this as Draft for now until we land #1040

rebornsick:
I removed 4 labels under AddonsConfig.jsx that seems not in use, and changed the order of "Pin LED" to "LED Pin".

PR #1041, opened 2024-06-04
Enables usage of protobuf enums in web config

Pelsin:
Compiles enums.proto into typescript defs.

Requires an extra build step, has been added to run before `npm run dev` & `npm run build` so it should work.

Example usage in Pins.ts
Screenshot 2024-06-04 at 13 33 06

Bundle size grew but not by a ton, still ok i hope!
Screenshot 2024-06-04 at 13 33 50


Pelsin:
Because of nanopb issues this not might be the way to go, ill close the PR and reiterate

PR #1042, opened 2024-06-04
Update to the images used for capture and home in the input history

TheTrainGoes:
The icons for the home and capture buttons in the input history on Switch mode were swapped.  This RP swaps them back.


arntsonl (approved):
Looks good

PR #1048, opened 2024-06-08
Add initial Japanese Translation for Web Config

mizma:
My initial draft for Japanese localization of Web Config.
I would like some feedback/review if someone can help with proof reading.

I have refrained from adding/changing the site code itself for any of the missing translation strings.


mizma:
One question on Tilt Input Add-on

I am unsure of this feature detail and undecided how to translate.
I'm reading from the doc that some input in specified GPIO pin activates some ammount of analog joystick input to some axis.  what I'm not sure is the relationship between Tilt (1|2) Pin and the Tilt (Left|Right) Analog (Up|Down|Left|Right) Pin.
does this mean both Tilt 1 Pin and Tilt Left/Right Analog ??? Pin needs to activate to send analog joystic?

mizma:
OK, I think I have figured out all available localization strings that seem relevant to be translated.

arntsonl (approved):
Looks fantastic, verified kanji and syntax is correct. Thank you!!

PR #1050, opened 2024-06-09
Update to the Zero_Rhythm board config

TheTrainGoes:
This board config update sets the default display to the custom board defined display now that the LED issue has been resolved.


arntsonl (approved):
Looks good

PR #1051, opened 2024-06-09
Adds configuration for the Flatbox Rev5 USB Passthrough

TheTrainGoes:
This PR adds configuration for the Flatbox Rev5 USB Passthrough


arntsonl (approved):
Yessss PS5 compatible flatbox rev 5 here we go

PR #1054, opened 2024-06-10
don't initialize the USB host if no one is listening

bsstephan:
Found this while looking into a different issue. if there's no drivers/addons that have registered as a USB host listener, there's no point in initializing it, so don't.

I wonder if we should also only underclock ourselves in this block too? (Not that I think it matters much.)


arntsonl:
Fixed in PR#1055 / added from this PR. Thank you @bsstephan !!

PR #1055, opened 2024-06-11
Fix for USB host causing lag on force reboot

arntsonl:
Potential fix for USB host causing reboot to lag out


arntsonl:
Thanks @bsstephan for the listeners.size() > 0

bsstephan:
Looks great to me, resolves all the recent regressions I have noticed.

arntsonl (commented):
Should be good

PR #1059, opened 2024-06-13
Configurable Debounce In BoardConfig.h

speedypotato:


arntsonl (approved):
This makes sense as a first support for optical!

PR #1060, opened 2024-06-14
use touchpad + dpad to emulate tpad click position on PS4

### LINKED ISSUE #924: PS Touchpad coordinates on button presses as actions
Currently, our touchpad implementation sends the button state when pressed, but does not send any positioning data, opting to set the state as untouched.  It's unknown whether this would cause undesired behavior in some games, but for current uses, there may be value including some sort of positioning data when the button is pressed.  A known implementation of this defines regions for the position coordinates based on the state of the D-Pad, but that could pose some additional behavior defects.  For example:

* D-Pad Neutral + TP = Center of Touchpad
* D-Pad Left + TP = Left of Touchpad
* D-Pad Right + TP = Right of Touchpad

A proposed means of implementing this would be to set different "actions" that could be bound to button presses instead, avoiding the need for a hotkey or macro dependency.

* Neutral - Current Touchpad behavior
* Left
* Right
* Center

bsstephan:
when the TP "button" is pressed, we formerly did nothing with the touchpad data, including not even sending the bit to indicate there is touchpad data, which was fine for normal purposes. however, we can use the dpad to emulate position data as if a finger was actually touching it, and send touchpad data for the eight directions. this will allow us to simulate clicking the left vs. right sides of the touchpad, which will satisfy the PS1-on-PS4 emulation of the Start and Select buttons.

this does *not* expose the different position clicks as an action or anything like that, hoping that this behavior comes with no real downside, since (AFAIK) touchpad click position data isn't used all that often on PS4/5 games.

Resolves #924


arntsonl (approved):
Super cool, thank you! Will be nice to have the "mystery" section of PS4 input reports filled out correctly.

PR #1061, opened 2024-06-15
Add a generic USB HID mode

bsstephan:
All the existing HID modes have quirks and host-specific decisions --- specific VID/PIDs, how the analogs are ordered, weird counter stuff, undocumented analog button payloads, auth packets, etc. Meanwhile, Bounty 35 has requested additional inputs for USB HID devices, for people who want to map more than 14 buttons. That bounty also (rightly IMO) surmises that we'd have to break out a new mode. The goal might be one day that this new mode can be user-configurable with specific button counts or who knows what else, especially once we get I/O expanders integrated --- maybe you could define a bunch of analogs!

Anyway, this PR:
1. copies the old USB HID mode into a "PS3" mode, since it has a couple PS3-specific quirks for the PS button and analog buttons
2. starts a new Generic HID mode, largely from scratch, but based on what is now PS3 mode
3. adds 12 "Extra" buttons, which only work in Generic HID mode, bringing Generic HID mode to:
4. one dpad hat, two emulated analog sticks, and 32 buttons (two of which don't do anything due to current lack of use in the Gamepad code)
5. exposes everything necessary to map them in the webconfig Pin Mapping page


arntsonl (approved):
Looks great!!! New Generic mode makes it much easier to split support, and now we can have our PS3 quirks and our generic USBHID

PR #1068, opened 2024-06-17
Re-add the copy base profile functionality

Pelsin:
Screenshot 2024-06-18 at 00 11 03


arntsonl (approved):
I'm glad that was easy to add back in :)

PR #1070, opened 2024-06-18
Hub Support!

arntsonl:
Finally got hub support working!

The big issue was first having an outdated TinyUSB which I fixed in 0.7.9. This caused a MASSIVE amount of issues and was the major blocker for 0.7.9 release.

Next issue was keeping track of "dev address" and "instance" inside of each listener.

Final issue was taking care of figuring out if we're a keyboard, dongle, or xbox one dongle. This code could use improvements but works well for now.


arntsonl (commented):
Hub incoming!!

mytokat8484:
whats hub? what does it do?

PR #1071, opened 2024-06-18
MiSTercade V2 Config

misteraddons:
New version of my MiSTer FPGA JAMMA adapter, now powered by GP2040-CE! One RP2040 IC per player, connected to an onboard USB hub.

This board takes advantage of the new expanded HID Direct input mode:
Joy2 and Start 2 are connected to player 1's RP2040-CE as extra buttons. This allows for twin stick games on a 2 player arcade cabinet (Smash TV, Total Carnage, Robotron 2084, etc.)

I modified the USB PID to match my VID:PID, but reverted it before merging to not cause issues.


arntsonl (approved):
Looks good! I'll update the png paths once this is in

PR #1072, opened 2024-06-19
I2C Autodetection

mikepparks:
In an effort to reduce the amount of work needed for migration and configuration of I2C addons, this PR aims to allow all I2C addons to be autodetected on startup.  This removes the need for I2C address and block fields in webconfig and subsequent configuration data.  A device only needs to be included on a previously-enabled I2C peripheral block and its addon enabled.

* When ``available()`` is called, if the addon is enabled, the list of known I2C addresses for the device will be passed to the ``PeripheralManager`` to be scanned.
* If a device is found at the address, the addon currently treats it as available and enables itself. Otherwise, the addon is disabled, and does not get included in the list.
* This currently does not include whether or not the device is _compatible_ with the addon, only whether it exists at the address.  Further detection would likely need to be done prior to inclusion in the update loop.

This approach does allow the autodetection of supported Display devices.  However, further work will need to be done to either allow interoperability between the Display and other devices sharing the I2C block, or a "priority" mode where the block cannot be shared with others.


arntsonl (approved):
Looks great! Very good feature to have with our i2c peripherals

arntsonl (approved):

PR #1075, opened 2024-06-19
Add board flash 0x9f SPI command to web config

arntsonl:
This adds the 0x9f SPI identification command to the web config memory info.

The system stats now show the actual board flash size identified by the flash chip itself.


arntsonl:
This crashes the board randomly :(

Before I close this as not an option, I am going to try to add the SPI flash cmd to the beginning of main() and see if we can consistently pull that at the beginning.

It's a neat feature but not worth crashing web config.

arntsonl:
A MUCH better idea is to do this in the gp2040 setup() or before that somewhere we know will not crash

TO-DO:

add this code to setup()

    uint8_t txbuf[FLASH_STORAGE_TOTAL_BYTES] = {0};
    uint8_t rxbuf[FLASH_STORAGE_TOTAL_BYTES] = {0};
    txbuf[0] = FLASH_STORAGE_CMD;
    flash_do_cmd(txbuf, rxbuf, FLASH_STORAGE_TOTAL_BYTES);

then store the value into a permanent location so it only happens once!

PR #1078, opened 2024-06-22
Update i18n localization strings

mizma:
Added few major non-translated strings into localization string.
Pruned some de-referenced localization string (to aid in future localization effort).
Added JP translation to the added localization strings from recent update + this update.

p.s. I did search and made sure they are not referenced but if some pruned strings need to be put back in. pls let me know so I can revert the change.


arntsonl (approved):
Good clean-ups! Lots of old text in there, good we keep that stuff removed and updated.

SavageCore:
Thanks @mizma

PR #1088, opened 2024-07-01
Enables choosing turbo and macro in pin mapping

Pelsin:
`GPIO pin to trigger 2+ buttons` Was started before these options could be done in the pin mapping, and was missed on merge.

Now selectable again
Screenshot 2024-07-01 at 19 38 11

Thanks to Hylian for finding it!


arntsonl (approved):
Oh nice, yeah that's good to have in there

PR #1090, opened 2024-07-09
DRV8833-Based Rumble (for XInput)

nicolasrmerz:
Adds DRV8833-based rumble functionality for XInput.

There are three main sets of changes:
1. Added a GamepadRumbleState struct, currently containing left/right motor rumble and unused stubs for left/right trigger rumble for XBone.
2. Fix some oddities with USB out endpoint writes in XInputDriver.cpp so that both LED and Rumble data is handled. Added a function in XInputDriver.cpp that checks if data in the out endpoint is rumble data and writes to GamepadRumbleState if it is.  (Sidenote: I know there's a proper GamepadAuxState WIP - I believe the writes to GamepadRumbleState can be trivially ported to GamepadAuxState when it is fully merged).
3. Added a DRV8833 Rumble Addon that reads from GamepadRumbleState and sends PWM to webconfig-defined pins that are hooked up to the DRV8833 module to start the left/right motors. There is also duty cycle rescaling to set the range of duty percentages that the addon will use (though 0 will be an "off" signal no matter what).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,7 @@ src/addons/playernum.cpp
 src/addons/playerleds.cpp
 src/addons/rotaryencoder.cpp
 src/addons/reverse.cpp
+src/addons/drv8833_rumble.cpp
 src/addons/turbo.cpp
 src/addons/slider_socd.cpp
 src/addons/wiiext.cpp
--- a/headers/addons/drv8833_rumble.h
+++ b/headers/addons/drv8833_rumble.h
@@ -26,11 +26,21 @@
 #define PWM_FREQUENCY 10000 // 10 kHz
 #endif
 
+#ifndef RUMBLE_DUTY_MIN
+#define RUMBLE_DUTY_MIN 20.0f
+#endif
+
+#ifndef RUMBLE_DUTY_MAX
+#define RUMBLE_DUTY_MAX 100.0f
+#endif
+
 // DRV8833 Rumble Module
 #define DRV8833RumbleName "DRV8833Rumble"
 
-// Scale uint8 to 0 - 100 range
+// Scale uint8 to 0 -> 100 range
 #define motorToDuty(m) (100.0f * (m/255.0f))
+// Rescale from 0 -> 100 range to min -> max range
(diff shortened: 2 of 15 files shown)


mikepparks:
After getting my own DRV8833 module to test with, here's some things worth noting for documentation's sake.

* Motor sleep pin setting can be optional (set to -1) to free up IO. Bring the module's SLP high by wiring to VCC. Sleep pin is likely preferred for the sake of power savings.
* When setting pins for AIN#/BIN#, the pin that will not be controlled should be wired to GND.
* Has been tested with [small disc motors](https://www.adafruit.com/product/1201) as well.
* May need to press a button on the controller first before rumble becomes active.

nicolasrmerz:
Further notes.

PWM frequency set to 10 KHz by default. Probably doesn't need to be ever touched by the average user, but wanted to leave customizable. With the DRV8833, anything under 50 KHz should fly.

### Duty cycle min/max
First, consider Duty Cycle Min = 0 and Duty Cycle Max = 100 (ie. no rescaling done). The XInput rumble packets and the GamepadRumbleState structs use uint8 values for rumble, giving a range of 0-255. This range gets translated to duty cycle factors of 0/255, 1/255, 2/255, ... , 254/255, 255/255, giving a new range of 0% - 100%.

The problem is that the lower end of this range gives too low an average voltage to drive the motor (the minimum operating voltage is motor-specific). The Duty Cycle Minimum/Maximum seek to resolve this. Setting these rescales the 0% - 100% range to DutyCycleMinimum% - DutyCycleMaximum%. Note, however, that no matter the rescaling, uint8 value of 0 will ALWAYS turn off the motor.

Experimentally, with the XBOne motors (Vybronics VJQ24-35F580C), the motors started operating reliably at about 25% duty cycle, meaning I set Duty Cycle Minimum to 25% and kept Maximum at 100%. Thus, for the nonzero uint8 range of 1 - 255, the voltage will range from about 5V * 0.25 = 1.25V to 5V * 1.00 = 5V.

These values will have to be obtained on a per-motor basis by the user either from the motor's datasheet or experimentally.

arntsonl (approved):
Fantastic work, thank you for contributing this!

PR #1099, opened 2024-07-31
Updates Flatbox Rev.5 Config & Header Documentation

brettohland:
This PR does the following:

- Adds the missing column on BoardConfig.h which correlates the screen printed switch labels with their GPIO pins
  - Re-orders the rows to be ordered by the screen printed labels
- Adds the screen printed labels, GPIO, and Input information to the Readme.md file


arntsonl (dismissed):
I'm okay with these changes, I see how they reflect the PCB labeling vs. the config headers

arntsonl:
I like the way these look based on the PCB labeling. We can revert it if we need to, but it looks clean to my eyes

PR #1102, opened 2024-08-05
New addon to support controller input from usb host

smahdink:
A new addon is added to the gp2040-ce named ControllerHostAddon, that enables getting input from USB controllers through the added USB host port in PICO and USB Host addon.

**Currently supported controllers are:**

1. Google Stadia controller
2. Sony Dualshock 4 controller


smnagavi:
Thank you for the notes, can you elaborate about your first point a little more?
About the name I actually like the name you suggested and if everyone agrees I would change the name.

smahdink:
Thank you for the notes, can you elaborate about your first point a little more?
About the name I actually like the name you suggested and if everyone agrees I would change the name.

mikepparks:
Lines 16-19 in ``headers/addons/controller_host_listener.h`` for example. I know why those bitfields are unnamed, but for the sake of having code that can be read & understood, we'd prefer if they were named.

mikepparks:
It appears that this is causing main Gamepad inputs to not function when enabled.  Your ``process()`` function does not check for changes in your own gamepad state, which continuously overwrites the Gamepad's values with your own. This prevents core functionality from occurring, including being able to use shortcuts on boot to load into bootsel or webconfig.  I would say to only update Gamepad state when:

* A supported device is connected to USB Host
* The input state from that device has changed since its previous iteration
* If the device is disconnected, no changes are made to Gamepad.

This will need to be addressed before we can determine if it's functional enough to merge.

arntsonl:
We're going to push this into the next version of the release. We want functionality like this, but this does open a big can of worms. I think moving our keyboard/mouse/gamepad host into a single host controller with host drivers similar to how we do drivers now is going to be the magic.

arntsonl:
Reimplemented in PR#1221!

PR #1104, opened 2024-08-08
Reactive LEDs

mikepparks:
This adds a basic implementation of reactive lighting using standard, non-addressable LEDs that react to ``GpioAction`` settings. Each LED has a setting for Default (not pressed) and Active (pressed) modes.

* Static On
* Static Off
* Fade In
* Fade Out


arntsonl (approved):
Super cool!! We might consider pulling this add-on into our LED code when we revamp that, but this is a great feature

PR #1107, opened 2024-08-09
Add Insert and Tilde to keyboard mapping

henrebotha:
Tilde was actually already there, just labelled in a way people might not expect. This commit elaborates on the label for that keycode to make it a bit more obvious.

This addresses Bounty #41 - Add Tilde & Insert to keyboard mode.


arntsonl (approved):
Easy fix!

PR #1111, opened 2024-08-11
Add dpad directions to hotkeys

Pelsin:


arntsonl (approved):
Neat, that makes sense to me

PR #1113, opened 2024-08-12
New Display screen modes: Pin Viewer and Stats

mikepparks:
Added two new Display screen modes and some minor adjustments to accommodate those changes.

Pin Viewer
* Displays the pin number and ``GpioAction`` definition if one exists.
* If a pin has no definition, or is assigned/reserved to a non-input pin, it will be listed in a separate section.

Stats
* Displays the version, build, type of build, board type, and board architecture of the current firmware.


arntsonl (approved):
very cool little useful screens

PR #1114, opened 2024-08-12
Fix vulnerabilities

henrebotha:
This is the result of running `npm audit fix --force`.

### DIFF
--- a/www/package-lock.json
+++ b/www/package-lock.json
@@ -42,7 +42,7 @@
 				"eslint-plugin-react-hooks": "^4.6.0",
 				"eslint-plugin-react-refresh": "^0.4.1",
 				"express": "^4.18.2",
-				"nodemon": "^2.0.22",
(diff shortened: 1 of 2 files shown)


Pelsin:
This does not only update the affected dependencies(nodemon) but also reruns an install ignoring earlier package-lock.json. Re-installing all dependencies according to package.json.

Thats why it update all packages, e.g "vite" in package-lock.json to 4.3.9 -> 4.5.3, which is fine as long as we are aware and have tested these updates.

henrebotha:
Ah good call, I'll see if there's a way to make it respect the lockfile.

arntsonl:
I'm going to set this back to draft until we can verify all the vite changes are valid.

arntsonl:
Yeah I'm going to close this one up. We can do a bigger push to get versions updated, nodemon is not critical right now.

PR #1119, opened 2024-08-18
Add axis specific last win SOCD cleaning

nevius:
It's a feature my Punk Workshop Leverless had before I changed the PCB to a RP2040. 

I think it's quite useful for certain games. For example in Street Fighter doing fast dive kicks with Cammy, I would often miss the DOWN input on the QCB, because my thumb as still on UP. With Y-Axis Last Win, the problem can be solved while still allowing for SOCD dashing on the X-Axis.

Is this something you would be willing to include in the main branch?

**TODO**
- Translations for JP / CN / BR are still missing


www/src/Addons/Tilt.tsx
```diff
@@ -86,7 +86,7 @@ export const tiltScheme = {
 	tiltSOCDMode: yup
 		.number()
 		.label('Tilt SOCE Mode')
-		.validateSelectionWhenValue('TiltInputEnabled', SOCD_MODES),
+		.validateSelectionWhenValue('TiltInputEnabled', TILT_SOCD_MODES),
```
nevius, line 89: This looked wrong to me, so I changed it.

src/display/ui/screens/ButtonLayoutScreen.cpp
```diff
@@ -186,11 +186,13 @@ void ButtonLayoutScreen::generateHeader() {
...
-		case SOCD_MODE_NEUTRAL:               statusBar += " SOCD-N"; break;
-		case SOCD_MODE_UP_PRIORITY:           statusBar += " SOCD-U"; break;
-		case SOCD_MODE_SECOND_INPUT_PRIORITY: statusBar += " SOCD-L"; break;
-		case SOCD_MODE_FIRST_INPUT_PRIORITY:  statusBar += " SOCD-F"; break;
-		case SOCD_MODE_BYPASS:                statusBar += " SOCD-X"; break;
+		case SOCD_MODE_NEUTRAL:                         statusBar += " SOCD-N"; break;
+		case SOCD_MODE_UP_PRIORITY:                     statusBar += " SOCD-U"; break;
+		case SOCD_MODE_SECOND_INPUT_PRIORITY:           statusBar += " SOCD-L"; break;
+		case SOCD_MODE_FIRST_INPUT_PRIORITY:            statusBar += " SOCD-F"; break;
+		case SOCD_MODE_BYPASS:                          statusBar += " SOCD-O"; break;
```
nevius, line 193: I changed this to SOCD-O because I wanted to use SOCD-X for the new one. Not sure if that's OK.

arntsonl:
For this one, we've been talking internally a bit and it might make more sense to have a per axis SOCD rather than the one-off combinational solutions. I see exactly what you're going for, but I think we're still deciding on how to go forward on this one. Thanks for the PR!!

nevius:
@arntsonl Thank you. Let me know if I can be of assistance.

mizma:
@nevius if you need JP translation, ping me via discord.

arntsonl:
Closing this as we are not going with single-direction SOCD at this time. If we come back to this I will reference the PR #

PR #1121, opened 2024-08-18
Duplicate USB report sending fix

mikepparks:
Removed ``tud_hid_report`` call in ``set_report`` to prevent unexpected behavior by rebroadcasting reports. Will need to test both input modes and USB Host addons for regressions.


arntsonl (approved):
Yup, this is definitely a relic of usb example

PR #1122, opened 2024-08-20
Add Analog Smoothing to thumbsticks

NickGuyver:
Add the ability to reduce jitter on thumbsticks by leveraging exponential moving average.


arntsonl (approved):
Nice and clean, thank you!

PR #1126, opened 2024-09-01
Add Toggle Joystick Add-on

tupleu:
Adds the Toggle Joystick Add-on which allows the user to switch dpad modes while held as well as change the tilt of the joysticks.

First, the dpad toggle buttons. When held, the dpad mode is switched to the selected dpad mode. When released, the dpad mode is switched back. Additionally, if any direction on the joystick is held, the input is frozen. For example the user can freeze the left joystick to hold up, while switching to the right joystick to move it around. This allows for a smoother experience when needing to utilize both the left and right joysticks.

The user can select two different buttons, a primary and a secondary toggle, both customizable in the add-ons configuration. For example, the primary can switch the dpad mode to right joystick and the secondary can switch to the dpad. The primary toggle will take precedence over the secondary toggle as well as freeze any inputs held by the default dpad mode and the secondary dpad mode. This allows for the simultaneous input of the left and right joysticks as well as the dpad.

Additionally, this add-on comes with a tilt toggle button that changes the intensity of the joysticks when held. The tilt factor can be changed in the add-ons configuration. This implementation allows for use of the dpad and meshes well with the rest of this add-on.

The pins for the primary, secondary and tilt toggle have been implemented in the pin mapping section, which allows for the buttons to be different for different profiles.


mikepparks:
So, how is your Tilt mode different than the other Tilt addon which does the same and is more robust?

tupleu:
> So, how is your Tilt mode different than the other Tilt addon which does the same and is more robust?

The other one doesn't allow for the use of the left and right joystick dpad modes. The direction pins have to be assigned individually and it doesn't allow for duplicate directions. The whole implementation is rather clunky seeing as you have to set the dpad mode to dpad, unmap most of the directions and then map certain buttons to the joysticks just to make it functional. It also has not transitioned to the pin mapping section which means that you cannot have it activated for one profile and deactivated for another profile.

The reason I decided to implement this version of the tilt functionality is because the other tilt add-on didn't have the functionality that I needed to work well with other add-ons (such as pin mapping). I don't think my implementation necessarily needs to replace the other tilt add-on, but it can be used in scenarios where the other one can not. My tilt mode implementation is much simpler as it only requires the assignment of one pin instead of multiple. This simplicity fits my use cases better than the other one. I didn't want to make any breaking changes to the other add-on, but maybe the two can be merged by someone more familiar with the other one's use cases.

TheTrainGoes:
@tupleu what is your use case for this?

tupleu:
@TheTrainGoes This add-on allows for a smoother experience when playing games that require dual analog sticks. For example being able to move and look around at the same time. For me, this feature has been a life saver for playing rpgs with the ability to look around. I can also quickly access the dpad buttons and switch back to the default left joystick with the press of a single button.

Here's a video of showcasing a similar implementation: https://youtu.be/A7U0vUZLelw?si=OAwyNAAd8irbpfjd

TheTrainGoes:
@tupleu we are not looking to have multiple Tilt style add-ons at this time.  

If you would like to merge this into the existing Tilt add-on we can take a look at it further.

tupleu:
@TheTrainGoes I have combined the two add-ons. Since the frozen inputs must be applied after the tilt, so that they do not also get tilted, the tilt and dpad mode toggle are implemented together. This reduces the chance that a code refactor could disrupt the order of these operations and cause a regression. Intuitively, both the tilt and dpad mode toggle function the same way as they are active when held and deactivate when released so it makes sense to have them together.

TheTrainGoes:
Thank you fo rthe combination @tupleu .  There are still some conflicts with main as outlined below.  If you can resolve those we can make a build artifact to test this out.

mikepparks:
You did not combine the two addons. You removed the entirety of the previous addon and deprecated its settings with no effort to migrate the previous' settings into your own.

tupleu:
@mikepparks This is why I didn't want to remove the other one in the first place. The other add-on is simply incompatible with other add-ons. It requires the remapping of several pins making them unusable for anything else and implements its own SOCD cleaning solution. I did try to replicate the settings of the other add-on by reimplementing them to be more compatible with the rest of the codebase. As for the migration of settings I can try to implement it, but I am not familiar with how this project handles migration and legacy code. Is all of the migration code contained in `src/config_utils.cpp`?

PR #1128, opened 2024-09-02
Make extra keys usable in keyboard mode

rmaccrimm:
This PR makes the 12 Extra keys added by #1061 usable in keyboard mode. I found that currently you can map these buttons to keyboard keys in the web config but they won't actually save.

My specific reason for doing this was to re-map the extra up button on my Mavercade controller to a unique training mode shortcut in Street Fighter 6.


arntsonl:
@rmaccrimm could you pull in latest and increment your config.proto enums based on @mikepparks 's changes? Thank you!

rmaccrimm:
> @rmaccrimm could you pull in latest and increment your config.proto enums based on @mikepparks 's changes? Thank you!

Sure, done!

arntsonl:
Thank you! Re-running the build on this and I'll get it merged.

arntsonl (approved):
Looks great! I think A3/A4 were missed in the keyboard driver on the other PR, so I'll put that in a different PR.

PR #1133, opened 2024-09-13
switch midpoint fix

NickGuyver:
Fix the bad midpoint, from 16 to 8 bit conversion, on analogs for Switch mode. Tested on Switch to be it's true center.


arntsonl (approved):
Sounds good, takes care of rounding errors when we have a maximum value at dead center.

PR #1136, opened 2024-09-16
Custom analog error rate

NickGuyver:
Add the ability to set a custom circularity error rate % for analog sticks.


headers/addons/analog.h
```diff
@@ -69,6 +69,10 @@
 #define SMOOTHING_FACTOR 5
 #endif
 
+#ifndef ANALOG_ERROR
+#define ANALOG_ERROR 0
```
Pelsin, line 73: If the default value is zero we need to modify form [validation](https://github.com/OpenStickCommunity/GP2040-CE/pull/1136/files#diff-1dc51191c1aed5b94ed7849f6d8458038a17ed0c47da979dd4de7e7adc3a8c99R106)
As zero does not exist as any value here: https://github.com/OpenStickCommunity/GP2040-CE/pull/1136/files#diff-1dc51191c1aed5b94ed7849f6d8458038a17ed0c47da979dd4de7e7adc3a8c99R25

Otherwise we will get the error:
Screenshot 2024-09-16 at 22 09 03

When enabling switch the first time.

Pelsin:
After saving with the value `9`, getAddonsOptions returns 
Screenshot 2024-09-16 at 22 15 18
Meaning it wont match any of the options and show the error:
Screenshot 2024-09-16 at 22 17 20

arntsonl (approved):
Looks great! More stuff converted to tsx is better

PR #1137, opened 2024-09-16
Minor fixes

Pelsin:
Validation same as TURBO_SHOT_MIN in https://github.com/OpenStickCommunity/GP2040-CE/blob/main/src/addons/turbo.cpp#L12


arntsonl (approved):
Clean fix!

PR #1146, opened 2024-09-24
Core0/Core1 Sync USB Stall fix

arntsonl:
Quick fix for getting auth back into PS5. Simple change


arntsonl:
Or I rewrote ps4 auth entirely.... same same

PR #1148, opened 2024-09-26
Xbox 360 Fix

arntsonl:
Fix for when descriptor is called repeatedly for xbox 360


arntsonl:
This fixes some stability issues I was having and also makes host

Issue #1140, opened 2024-09-17
Add an enabled field to profiles, to not allow switching to them (esp. useful when cycling)


bsstephan:
From the roadmap in #1123 --- being able to disable profiles would make it possible for the cycle hotkey/button to skip profiles that aren't being used, making that action more convenient, especially if/when we increase the number of profiles. This would be something that can be managed by the user in the webconfig, though I think we need a short conversation on if the default value should be enabled or disabled.

bsstephan:
This will be in v0.7.10.

PR #1149, opened 2024-09-26
Allow for the enabling and disabling of profiles

bsstephan:
This...
* adds an enabled flag to GpioMappings (so in both the core gamepad Profile 1, and the alternative Profiles 2-4 structures)
* updates the set/next/previousProfile calls to only set or cycle through enabled profiles; if you have 1 and 2 enabled and 3 and 4 are disabled or don't exist, next toggles 1 -> 2 -> 1, or activating the hotkey to set profile 4 just doesn't do anything (neither trigger a change, so the OLED doesn't even update --- the request is essentially silently ignored)
* allows the user to manage the enabled flag via the webconfig for profiles 2(!) through 4
    * profile 1 cannot be disabled and is always assumed to be enabled in the code
* because profiles no longer have to be populated/can be disabled, profile 1 is no longer copied into 2-4 when initializing a fresh board
    * there is now a migration to disable profiles that look like they are still copies of profile 1 (I guess in retrospect they could be deleted, but whatever)
* because profiles 2-4 may not exist, the webconfig also allows the user to create them for the first time, based on profile 1, where they can then populate them

I have played with the hotkeys to select (or not) profiles heavily, and also have tested the webconfig on a migrated board, a reconfigured board, and a nuked board. Everything seems good. And thanks again to another awesome UI job by @Pelsin.


bsstephan:
This resolves #1140.

arntsonl (approved):
Looks great! Next/Previous makes sense, just loop through until you find a valid profile, which could be the profile you're currently on.

PR #1151, opened 2024-09-27
Add pico-sdk 2.0.0 support

Prime9999:
- Update Pico-PIO-USB to 0.6.1
- Notifying pico-sdk version to compile properly (e.g. Pico-PIO-USB)


arntsonl:
We'll pull this in after 0.7.10 RC1. We'll need to do a lot of testing with a new PIO version because it touches every part of our USB Host.

Fortinbra:
This looks great, could I get you to also make this change to the cmake.yml file? This will update the version of the SDK used by the CI pipeline.

`uses: Fortinbra/RaspberryPiPicoBuild@v4`

PR #1154, opened 2024-09-27
Fix initialization of new profiles' blocked pins and counts

bsstephan:
Fixes two fatal bugs:
1. Creating a totally new profile would not have any pins on a subsequent load of the board, making using the profile lead to an inoperable controller, because the pin count wasn't retained in protobuf.
2. Creating a totally new profile would not retain the `RESERVED` or `ASSIGNED_TO_ADDON` actions, because the old code assumed other methods were populating those profiles' values, but that is not possible on a totally new profile.


bsstephan:
Resolves #1155.

Pelsin:
Tested, works perfectly!

arntsonl (approved):

PR #1160, opened 2024-09-29
Helper component for debug stdio output on arbitrary GPIO pins

DvdGiessen:
Use case: While getting GP2040-CE to run on a prebuilt board, I ran into the limitation that I did not have access to any of the SWD or UART pins, making it hard to debug. However, there were a few other GPIO pins easily accessible.

This PR adds a small PIO component that allows outputting stdio to any GPIO pin, making use of [the UART TX PIO example code](https://github.com/raspberrypi/pico-examples/tree/master/pio/uart_tx) from Raspberry Pi.

It's not used by default, but can be easily set up by defining `PIO_UART_TX_STDIO_PIN` and calling `pio_uart_tx_stdio_setup()` to enable the functionality.

Not sure if useful, so sharing it as a draft PR.


mikepparks:
Given that PIO is a limited resource, it's likely that this will not be accepted as the PIO resources are better reserved for non-debug functionality.

The UART issue we're facing is merely a configuration issue where our build process needs to be aware of changes to ``PICO_DEFAULT_UART_TX_PIN`` and ``PICO_DEFAULT_UART_RX_PIN`` respectively.  Our peripheral manager is designed around the expectation that we're using the standard GPIO mapping for peripheral "blocks" on the RP2040, and UART is eventually going to be included in that, so using PIO as a solution for something already solved in the hardware itself a no-go.

DvdGiessen:
That makes sense. I suspected it might indeed be possible to use the built-in UART on arbitrary pins, but I couldn't get it to work on my board, while this PIO solution did give me a working output.

If it is not useful to have this included as an optional debugging helper I can close the PR. :)

arntsonl:
I'm closing this up, thank you for the suggestion though!

PR #1161, opened 2024-09-29
Implement QMI8658 gyro/accelerometer support

DvdGiessen:
This implements basic support for the QMI8658 IMU, which is for example used on a number of off-the-shelf Waveshare RP2040 boards (I'm using [this one](https://www.waveshare.com/product/raspberry-pi/boards-kits/raspberry-pi-pico-cat/rp2040-lcd-0.99-b.htm), see my branch with the board config [here](https://github.com/OpenStickCommunity/GP2040-CE/compare/main...DvdGiessen:GP2040-CE:main)).

This is a very basic implementation, making use of the MIT-licensed code from Waveshare to implement communications with the IMU. It uses the GP2040 I2C abstraction, and I've added a simple configuration for it that for now only allows enabling/disabling it.

Sharing it here as a draft PR; in case I don't find the time to finish polishing this up with extra settings, calibration, etc it might hopefully still be useful to others. :)


mikepparks:
Glad to see someone taking advantage of the new Aux structures.  Given that we're in the middle of rolling a release, this will likely need to move off to the next version to give us time to get hardware to check it out.

* Good catch on the Relative sensor change. I added Relative fairly late in the dev work and didn't update those.
* I would suggest fixing the indentation issue you have for ``optional QMI8658Options qmi8658Options = 28;`` in ``config.proto`` to match the others.
* It's preferred that we avoid the old naming for I2C terms, preferring "target" or "device".
* Debug output functions such as printf should either be removed or excluded via a conditional define for their usage, as they can potentially impact performance.

DvdGiessen:
Thanks for the feedback!

> * I would suggest fixing the indentation issue you have for `optional QMI8658Options qmi8658Options = 28;` in `config.proto` to match the others.

Fixed! I didn't notice my editor used `.editorconfig` to switch to tabs here; already had to fight it a bit for example due to the trim trailing whitespace option resulting in a lot of unrelated whitespace changes I had to exclude from my commits. :) Might make for another small but useful PR similar to #436.

> * It's preferred that we avoid the old naming for I2C terms, preferring "target" or "device".
> * Debug output functions such as printf should either be removed or excluded via a conditional define for their usage, as they can potentially impact performance.

I copied the QMI8658 library directly from Waveshare, so I didn't look into cleaning these up yet. I've addressed these two points for now; there's more things in these files that could be cleaned up but at least for now it is somewhat functional.

arntsonl:
This is a cool add-on, @DvdGiessen what do you think it would take to get this one done? I like the functionality!

PastorL69:
I'd also be interested in using an accelerometer, the BMI160 to be specific. 

There are many types, but the BMI160 seems to be quite a popular one. Maybe I'll have a look into it if something like a generic accelerometer/gyro implementation isn't available for the next release.

DvdGiessen:
I haven't yet made time to continue work on this, but have not forgotten about it.

> what do you think it would take to get this one done?

Aside from cleaning up the code to be easier to understand/maintain main thing would be adding a bunch of well thought out configurable options for calibration, smoothing, all kinds of things you'd want to build a good gyro control.

I suspect we'll need this because there isn't a lot of options to tweak this in most software that supports gyro/accelerometers; for example I tried with the usually excellent Steam Input for remapping but support there wasn't as complete as I had hoped. I suspect that since we're emulating a PlayStation controller there aren't that many variants of that, most software assumes they already have a fairly consistent calibration, respond in a specific way, etc. The code in this PR gives output, but is not very easy to use because of lacking calibration.

And important to keep in mind there would be bringing these improvements to a few other gyro chips as well. Aka not implement them in the QMI8658 addon but in a generic / reusable IMU addon; so we can more easily support multiple gyro's (such as the mentioned BMI160; I also got asked on Discord about the BNO086, LSM6DSO, and MPU6886).

Once this is implemented it would be very helpful to document it well. Recommending a few pieces of software to use it with if you want for example mouse emulation, mapping it to other control schemes, guides on how to calibrate the sensors properly, etc. That would make the functionality easier to get started with, and I suspect would be very welcome to those interested in using this for accessibility purposes.

> I like the functionality!

> Maybe I'll have a look into it

Feel free to ping me if you want to collaborate on it! No promises about me actually being available, but I told myself I should make some time for this. :)

Petruzzo73:
Any chance to have gyro/accel implemented? Thanks!

mikepparks:
@Petruzzo73 Gyro/accel is already implemented for PS4/5. This PR is for driver support, and is awaiting some upstream design changes to make that easier.

Petruzzo73:
Thank you. You mean that QMI8658 accelerometer can already be used for PS4/PS5?

Petruzzo73:
@mikepparks any update on this?

Thanks

arntsonl:
Going to close this PR, if we want QMI8658 gyro support, lets add a new PR and make it up to date.

PR #1167, opened 2024-09-30
only load a profile if profileNum exists and is enabled

bsstephan:
This handles both the gamepad profile number being a number that doesn't exist (even though that shouldn't be possible via webconfig), or if the profile is disabled (which could happen if the user picks profile N and then disables it afterwards). In both cases, this doesn't load the `alts` pins, so the functional pins come from profile 1.


arntsonl (approved):
Yep makes sense

PR #1168, opened 2024-09-30
PS Feature reports causing unexpected behavior on startup

mikepparks:
This should address the issue where the PS features were unexpectedly enabling the lightbar on startup, as well as a number of other incorrect values in the feature report.


arntsonl (approved):
Nice!

PR #1169, opened 2024-09-30
Update ConfigScreen.cpp

TheTrainGoes:
Small update to change the shortcut buttons for `pin viewer` and `stats` to buttons that will not conflict with the shortcut to get into web-config mode.


arntsonl (approved):
Nice, good to catch that now

PR #1177, opened 2024-10-01
Keyboard Host Clean-up / Unmount Bug Fix

arntsonl:
Small bits of clean-up to keyboard host. Should guarantee dev addr, instance, protocol type, mount/unmounts, and update loops


arntsonl:
I'm going to leave this for 0.7.11 as I don't think it impacts anything

PR #1178, opened 2024-10-01
Restrict runtime saves

### LINKED ISSUE #1176: Conflicts between saving to the flash while the USB host is active
It is unsupported to write to flash from one core while the other is using it. Our runtime flash users are the configuration and tinyusb. The current implementation has `writeToFlash` use `multicore_lockout` before writing to essentially disable interrupts on the other core (where tinyusb is running). This keeps from crashing the USB host, but the disabling of interrupts does cause e.g. a MagicBoots to unpair and thus authentication can be dropped if that's in the middle of happening when interrupts are disabled.

Meanwhile, we save the config to flash in the following instances:
* in Turbo when changing the shot count
* in NeoPicoLED (?? need to look into this one)
* `ConfigManager::setGamepadOptions` but this looks like vestigial code
* changing JS modes
* changing SOCD cleaner modes
* setting invert X/Y axis
* setting 4-way modes
* changing profiles
* changing input modes (at boot time)

Most of these are just saving to config after changing a runtime setting, e.g. the SOCD modes, profiles, and similar toggles. In order to not interrupt auth, I think not saving, or saving in a more limited/controlled manner, is worth investigating.

bsstephan:
Details in  #1176, but saving to flash and host USB conflict, with the current code making the flash write safe but disabling interrupts on host USB's core for long enough to cause the connected dongle to disconnect. This makes most of the saves only occur if host USB is disabled (though they are still forced in the webconfig), and add a hotkey to force a save if you know what you're doing.

Resolves  #1176

### DIFF
--- a/headers/storagemanager.h
+++ b/headers/storagemanager.h
@@ -49,6 +49,7 @@ public:
 
 	void init();
 	bool save();
+	bool save(const bool force);
 
 	// Perform saves that were enqueued from core1
 	void performEnqueuedSaves();
--- a/proto/enums.proto
+++ b/proto/enums.proto
@@ -300,6 +300,7 @@ enum GamepadHotkey
     HOTKEY_DPAD_LEFT             = 40;
     HOTKEY_DPAD_RIGHT            = 41;
     HOTKEY_PREVIOUS_PROFILE      = 42;
+    HOTKEY_SAVE_CONFIG           = 43;
 }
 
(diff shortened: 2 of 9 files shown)


arntsonl:
This should work well to prevent host disconnect. We'll review the user experience on this one, but I imagine for now we can do something like this.

arntsonl (approved):
Yeah let's add this in, and we can review user experience changes. This prevents auth from ever disconnecting

PR #1179, opened 2024-10-01
Updating pico build version

Fortinbra:
Updating the build system to use the pico-sdk 2.0.


mikepparks:
As of 2d16215, the build does load up on a normal Pico, and inputs register in Xinput mode. Trying to boot into webconfig results in a crash.

arntsonl (approved):
Nice, pushes us up to 2.0 SDK!

alt-0191:
Maybe the development documention show be update

PR #1188, opened 2024-10-04
INIT Pull Request Template

j-zhao:
### PROBLEM
Pull Requests descriptions are variable and inconsistent, depending on the contributor.

### SOLUTION
Init a GitHub PR template to help promote consistency and make it easier for both contributors and reviewers to understand intent.

Similar to Issues templates, Pull Request templates can be helpful to make PR contributions more clear and capture context. See tagged links for more information with how these integrate with GitHub

Feel free to propose adjustments this template's content

### NOTES

This is purely a suggestion, no pressure to adopt this. I find it to be very valuable when there are a large number of different contributors.

Here's how this would look:

image
image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,31 @@
+## <Issue-Link>: <A brief summary of this merge request>
+
+### PROBLEM
+<!---
+Outline the problem being solved by this code change.
+-->
+
+### SOLUTION
+<!---
+Outline how this Pull Request addresses the problem.
+-->
+
+### NOTES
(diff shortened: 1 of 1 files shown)


arntsonl:
This would be more in line with a work approach. We're for-fun, so I'm going to keep free form in PRs. If it becomes a problem we'll address it.

j-zhao:
> This would be more in line with a work approach. We're for-fun, so I'm going to keep free form in PRs. If it becomes a problem we'll address it.

Old habits die hard :)

PR #1191, opened 2024-10-09
Fix bare URLs in Markdown files to ensure they are displayed properly

tylervz:
Wrap bare URLs with angle brackets in Markdown files.

Rationale: Without angle brackets, a bare URL or email isn't converted into a link by some Markdown parsers.

https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md034---bare-url-used


arntsonl (approved):
Looks good!

PR #1192, opened 2024-10-09
Reformat JavaScript and TypeScript files with Prettier

tylervz:
I updated Prettier to the latest version and then I ran `npm run format` (which in turn calls `prettier --write "{src,server}/**/*.{tsx,ts,jsx,js,json,scss}"`).


arntsonl (approved):
Nice, gets us updated to the latest Prettier and updates a few fiiles

PR #1195, opened 2024-10-13
Move SOCD settings to profiles

Pelsin:
This PR only contains web ui changes, if someone wants to continue then this is a good start.

Adds `socdMode` to `setProfileOptions` and `setPinMappings`

Screenshot 2024-10-14 at 00 27 24


arntsonl:
This will be coming in a future PR, thanks @Pelsin !

PR #1204, opened 2024-10-27
update ko-KR localization strings

losernator:
* Correct Korean Flag and Polishing Korean Translations


arntsonl (approved):
Thank you for the updates!!

PR #1208, opened 2024-11-04
Added additional ways to get into BootSel mode

TheTrainGoes:
This PR adds two additional ways to get into BootSel mode for instances where you may not have a mapped S2 button or access to a mapped S2 button for a web-config BootSel reboot.


arntsonl:
We're going to axe this one in favor of other failsafes. But a good brainstorm

PR #1216, opened 2024-11-20
Standalone Directional Input actions and Addon buttons as GpioAction

mikepparks:
First, this introduces several new GpioAction-based directional input values.  These are intended to be standalone, meaning that they will work independently of their previous implementations which are based on either D-Pad Mode or DDI mode options.

* `D-Pad Up`
* `D-Pad Down`
* `D-Pad Left`
* `D-Pad Right`
* `Left Analog X+`
* `Left Analog X-`
* `Left Analog Y+`
* `Left Analog Y-`
* `Right Analog X+`
* `Right Analog X-`
* `Right Analog Y+`
* `Right Analog Y-`

Second, a number of Addons previously allocated pins for button/switch definitions.  In order to streamline Addons, and allow more button functionality to be mapped into the Pin Mapper and Profiles, the following Addons have been updated.  In order to use their functionality, the respective Addon will still need to be enabled, but they no longer require fields specifically to map a button for their use.

* Input Reverse
* Tilt Input
  * All pins intended for buttons to emulate their respective Analog Stick direction now use the new standalone Left/Right Analog input actions
  * `Tilt 1` and `Tilt 2` button definitions have been relocated to new `Analog Stick Modifier Low` and `Analog Stick Modifier High` actions
* Focus Mode
  * Much like the SOCD switches being relocated to Actions in Pin Mapping, the `Focus Mode` pin is now a mappable Action


arntsonl (approved):
Looks great!!

PR #1219, opened 2024-11-23
Regression fix (thank you BRP)

arntsonl:
Add rumble and reactive LEDs back in, I missed these on my modifications for Xbox 360 auth support


arntsonl:
This got missed in the release, sorry!!

PR #1226, opened 2024-11-30
Added support for the Ultimarc Ultrastik 360 to the gamepad plugin

not-magic:
This allows using the Ultrastick 360 to be used as an analog input by connecting it via the usb host port.


arntsonl (approved):
Looks great! That's a nice easy add to give us support for the ultrastik 360, very cool!

mexiace:
This is great.  Any way to send to send maps to the U360 using this method?  Have always had an issue using a U360 as a pure joystick (L, R, D, U) while keeping the ability to send maps via USB (Hybrid mode).

PR #1235, opened 2024-12-14
X-Input Serial ID (Pico ID Based)

arntsonl:
Update the X-input Serial ID to use the Pico unique ID instead of a static value.

This might affect multiple USB controllers plugged into the same computer, but it might not


Jimizan:
without different GUID option (inside WEB CONFIGURATION) be a problem
randomly confuse the usb ports order
since one of the main uses and purchases of these encoders is for video games and 2 encoders are used...(2 players)...
they all have the same problem,...randomly the inputs of the 2 players are reversed in the program/game of use.
a big limit of the gp2040 with this use!

arntsonl:
> without different GUID option (inside WEB CONFIGURATION) be a problem randomly confuse the usb ports order since one of the main uses and purchases of these encoders is for video games and 2 encoders are used...(2 players)... they all have the same problem,...randomly the inputs of the 2 players are reversed in the program/game of use. a big limit of the gp2040 with this use!

We took a look at how Batocera generates the GUID, and it appears to be based off of the VID/PID and not the reported serial ID. This should hopefully help Windows, but does not necessarily fix other distros.

This SHOULD help with https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/usbcon/selective-suspend-in-usb-drivers-wdf.md software based usb suspension. I've also verified in https://www.nirsoft.net/utils/usb_devices_view.html that both serials show up independently on two Pico devices.

PR #1236, opened 2024-12-15
Button for Specific Tilt Value on Analog Sticks

BighomerYABOY:
Can't you just add a modifier like this?


mikepparks:
This completely misses the entirety of the work done in the Tilt addon, and adds unnecessary hard-coded percentages in the process. The Tilt percentage modifier does not need to be in the core input handler. It's handled by the addon itself.

PR #1246, opened 2024-12-27
Prepare a system stats global state for future "new version banner"

Pelsin:


arntsonl (approved):
Looks great! I like the new graphics for flash and heap

PR #1247, opened 2024-12-28
Event System Implementation

mikepparks:
(Re-implemented event system from previous PR #971 to be up to date with current codebase.)

The idea behind this event system is to implement system-wide notices when certain functions occur.  This will allow for more reactive features and potentially improve cross-feature communication, while cutting back on or eliminating duplicate functionality.

Examples of usage:
* On WebConfig restart, an event is triggered and the Display feature will change to a "Restarting" screen.
* With rotary encoders, the direction of travel, as well as the encoder ID is sent in an event that could be used for adjusting turbo rate, or switching profiles.

To add events, the following is needed:
* A new event enum value added to ``GPEventType`` in ``proto/enums.proto``
* An event header in ``headers/events``
* Event handlers can be registered with ``EventManager::getInstance().registerEventHandler(GP_EVENT_ID, GPEVENT_CALLBACK(this->handleEvent(event);));``
* Events can be fired with ``EventManager::getInstance().triggerEvent(new GPEventClass(...props...));``


src/addons/display.cpp
```diff
@@ -66,6 +66,8 @@ void DisplayAddon::setup() {
     }
     gpScreen = nullptr;
     updateDisplayScreen();
+
+    EventManager::getInstance().registerEventHandler(GP_EVENT_RESTART, GPEVENT_CALLBACK(this->handleSystemRestart(event);));
```
arntsonl, line 70: is that an extra ; on handleSystemRestart? or is that part of the macro?

arntsonl (approved):

PR #1255, opened 2025-01-03
UI tweaks & colors

Pelsin:
- Change body, nav, text colors to match the documentation site. Still more that could be done.
- Text now higher contrast, should be easier to read
- Fixed issue where the entire website got refreshed when pressing the logo
- Replaced the blurry logo (perhaps to large?)
- Change delete macro row to be the same as in hotkeys, no need for double clicking now (see image)
- Hotkeys and macros now have a tiny bit better grid (see image)
- Format settingsPage
- Mobile nav dropdown looking better

| Before  | After |
| ------------- | ------------- |
|  Screenshot 2025-01-03 at 23 15 17 | Screenshot 2025-01-03 at 22 53 25 |
| Screenshot 2025-01-03 at 23 15 00 | Screenshot 2025-01-03 at 23 29 01 |
| Screenshot 2025-01-03 at 23 32 38 | Screenshot 2025-01-03 at 23 32 17 |


arntsonl (approved):
Looks fantastic! 25kb for the logo is not bad at all, I think we still have plenty of room in the flash. We can see when its on a board :)

PR #1256, opened 2025-01-04
New pico sdk

Fortinbra:
This builds and works with gamepad and webconfig, but does not work with RGBLEDS


arntsonl:
Fixed on latest main, thank you for all the work!!

PR #1259, opened 2025-01-08
Update GitHub Actions runners specifying ubuntu-latest

Pelsin:
Updates the workflows to use a specific version runner images rather than the named "latest" to fix CI warnings


arntsonl (approved):
Nice, yeah that'll fix those warnings which are very annoying

PR #1260, opened 2025-01-08
Add GpioAction for sustaining 4/8 way mode

Pelsin:
Would be interested to know if there is another preferred to handle fourWayMode in `Gamepad`


headers/gamepad.h
```diff
@@ -188,6 +188,8 @@ class Gamepad {
 	GamepadButtonMapping *mapAnalogRSXPos;
 	GamepadButtonMapping *mapAnalogRSYNeg;
 	GamepadButtonMapping *mapAnalogRSYPos;
+	GamepadButtonMapping *map4WayMode;
```
arntsonl, line 191: Take a look at the enum "SUSTAIN_FOCUS_MODE" for how we want to do this.

What we want is a single GpioAction pin definition that allows a user to toggle their current 4-8-way mode to the other setting. So basically its a boolen-not that toggles whatever setting you have assigned to gamepad.

proto/enums.proto
```diff
@@ -261,6 +261,9 @@ enum GpioAction
 
     BUTTON_PRESS_INPUT_REVERSE   = 69;
     SUSTAIN_FOCUS_MODE           = 70;
+
+    SUSTAIN_4_WAY_MODE = 71;
```
arntsonl, line 265: And this one we can simplify down to SUSTAIN_4_8_WAY_MODE = 71; since we can use a single GpioAction::

src/gamepad.cpp
```diff
@@ -83,6 +83,8 @@ void Gamepad::setup()
 	mapAnalogRSXPos = new GamepadButtonMapping(ANALOG_DIRECTION_RS_X_POS);
 	mapAnalogRSYNeg = new GamepadButtonMapping(ANALOG_DIRECTION_RS_Y_NEG);
 	mapAnalogRSYPos = new GamepadButtonMapping(ANALOG_DIRECTION_RS_Y_POS);
+	map4WayMode     = new GamepadButtonMapping(SUSTAIN_4_WAY_MODE);
```
arntsonl, line 86: Same with this one, we can simplify down to a single map48WayMode =

src/gamepad.cpp
```diff
@@ -257,7 +263,7 @@ void Gamepad::process()
 	}
 
 	// 4-way before SOCD, might have better history without losing any coherent functionality
-	if (options.fourWayMode) {
+	if (fourWayMode) {
```
arntsonl, line 266: And because we're doing a toggle, we can just do options.fourWayMode ^ map48WayModeSwitchOn or something similar

src/gamepad.cpp
```diff
@@ -365,6 +371,10 @@ void Gamepad::read()
 	else if (values & mapButtonRS->pinMask)	activeDpadMode = DpadMode::DPAD_MODE_RIGHT_ANALOG;
 	else					activeDpadMode = options.dpadMode;
 
+	if (values & map4WayMode->pinMask) fourWayMode = true;
```
arntsonl, line 374: then this is the same, we just have one pin to do the toggle

arntsonl (changes requested):

arntsonl (approved):
This is exactly how I would implement it too, looks great!!

PR #1261, opened 2025-01-08
Add RP2040 Mini Breakout Board USB Passthrough

TheTrainGoes:
Add RP2040 Mini Breakout Board USB Passthrough board


arntsonl (approved):
Looks good

PR #1264, opened 2025-01-08
Xbox One Multiplayer Fix & Auth Bug Fix

arntsonl:
Xbox One randomization for serial ID


arntsonl:
This is going to be a WIP while I work on different solutions to get multiple XBONE devices to show up

arntsonl:
There were two major issues here:

#1 I was never initializing xboneDriverState so it was being assigned to random values. This was causing issues in the auth code. Setting it to a new NOT_READY state forces it to not be ready to auth until we are truly ready.

#2 the Xbox One driver logic says either it can request a device driver through GIP_DEVICE_DESCRIPTOR (0x04) OR it can request it through GIP_CMD_LED_ON (0x0A), so we have to catch both and then I check to make sure its waiting for descriptor.

This fixes it, verified working on Windows 11 and real Xbox One.

PR #1266, opened 2025-01-08
20250108   remove legacy devices

TheTrainGoes:
Removes some legacy devices and those we no longer want to support releases for.


TheTrainGoes:
Closing as this appears to have pulled in my updates from the Mini Passthrough board as well.

PR #1271, opened 2025-01-16
Enables keyboard to use multiple of the same options across buttons

Pelsin:
Also fixing a ton of type errors and remove unused code.
The validation for the keyboardMapper should no longer be necessary.


arntsonl (approved):
Nice! Good QoL changes and fixes

PR #1276, opened 2025-01-26
Add validation for duplicate hotkey combinations

Pelsin:
Validation for this until how/if we should handle different hotkeys with the same combination.
 
Screenshot 2025-01-27 at 00 43 12

- Also aligns all the content in the rows vertically, some inputs where previously not centered

Could be better ways to do this, but could not think of any.


arntsonl (approved):
Looks good! the errors[0] || errors[0].buttonMask was probably some old issue we had in the errors code.

PR #1282, opened 2025-02-02
Mapped Stadia Capture and Assistant button to touchpad right and left

smahdink:
Some games require touchpad left and right clicks to fully function. I added the two extra buttons available, to be mapped to A3 and A4 in GP2040CE.


TheTrainGoes:
My preference here would be that the default mapping for 18 stay as A2 since it is labeled as a capture button on the controller.  No issues having 19 added in though.  Maybe a way to easily swap between 18 being capture and trackpad fuctions?

smahdink:
I can change the 18 button to trigger both A2 and A4. Does Nintendo Switch mode have touchpad support? I mean do A3 and A4 do anything in the case of Switch?

PR #1285, opened 2025-02-17
Player Num Removal + Add-on Post Process Rewrite

arntsonl:
Rewrote the add-ons and drivers a bit. This is a much cleaner way to do it, and should hopefully fix the timing in Turbo as well as any other add-ons having issues with USB reporting vs. actual timing.


arntsonl:
Turbo works fine, but this PR gave me some time to think about how we want to handle post-process (usb report complete) requests in the add-ons.

Currently no add-ons will use it, but this exposes the USB report completion from all of our drivers to the add-on system. We can use this if any add-ons rely on the USB reports being sent to do "something".

PR #1303, opened 2025-03-15
Return of GP2040-CE boot logo

arntsonl:
Returning the GP2040-CE boot logo to our restart screen


mikepparks:
Good to go on my end. Gets rid of the old animations without removing the logo from the new menu utilities that use it.

PR #1304, opened 2025-03-16
Mini Menu and Button Layout Fixes

mikepparks:
* Added new board defined alternate button layouts
* Except for the Blank layout, hide layouts from selection if no elements are defined
* Fixed toggle issue with Turbo & Focus Modes
* Prevent Turbo mode from functioning if no pin is defined or the addon state is reenabled


arntsonl (approved):
Looks great! Gives the board uf2s some flexibility on user-defined board layouts.

PR #1315, opened 2025-03-21
Update for Haute42|COSMOX Lite and Ultra

TheTrainGoes:
Update for Haute42|COSMOX Lite and Ultra to give both a board defined screen layout


arntsonl (approved):
Looks good! matches my ultra

PR #1318, opened 2025-03-23
Case RGB Linkage Mode Flicker Fix

arntsonl:
Modified animation effects so they return a bool if animation was suc…cessful.

This is important to our linkage frame as we use the "pre-button press" frame to copy to our case linkage RGBs.

Minor clean-ups to code


arntsonl (commented):
Looked over everything, we good

PR #1319, opened 2025-03-24
Add config for Haute42|COSMOX X Series Analog

TheTrainGoes:


arntsonl (approved):
Looks good, gives us another Haute config for the X

PR #1321, opened 2025-03-24
Unknown UF2 URL Downloads Fix

arntsonl:
Fix for URL tag name on unknown downloads for UF2.

Github might have changed this from release.data.tag_name to release.tag_name at some point here


arntsonl (commented):
This was a legacy bug we didn't catch

PR #1324, opened 2025-03-26
Hotkeys Fix + Menu and Event Tidying

arntsonl:
Fix for hotkeys not allowing repeats.

Re-arrange on the main menu screen code.

Added virtual to GPEvent destructors to make sure we are calling all the way up the chain.


arntsonl (commented):
Everything looks right

PR #1333, opened 2025-04-02
Web-Config and Various Fixes for RC2

arntsonl:
A lot of improvements for RC2, including proper JSON_ARRAY_SIZE() inclusion for web-config, actual JSON_OBJECT_SIZE() calls for our web config requests

- WebApi.js and Http.ts fixes for Promise.all to prevent all of the EMPTY connections
- DisplayConfig.jsx fix for setDisplayOptions in preview mode

STILL having crashes for preview mode, but this fixes a good number of things

### DIFF
--- a/headers/gp2040.h
+++ b/headers/gp2040.h
@@ -77,7 +77,10 @@ private:
     // input mask, action
     std::map<uint32_t, int32_t> bootActions;
 
+    void checkSaveRebootState();
+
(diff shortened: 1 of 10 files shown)


arntsonl:
Replaced this with PR#1335 and PR#1336

Still doesn't fix the crash, but doesn't make sense to merge all of this as a single PR

PR #1334, opened 2025-04-03
Restructure fetching to not trigger abundant saves

Pelsin:


arntsonl (approved):
Looks awesome!! So much better without all the calls

PR #1336, opened 2025-04-03
Web Config Memory + Force Save -> Event Save Fixes

arntsonl:
Memory allocation fixes for our web config DynamicJsonDocument declar…ations.

Fixes for storage save -> EventManager save


arntsonl (commented):
Everything should work a bit smoother now in terms of memory profile and properly sizing our DynamicJsonDocument on the heap

PR #1338, opened 2025-04-05
Add a command to check for changes in en locale strings since last update of another language locale.

mizma:
This is a trial implementation to check what work needs to be done on Locale strings to catch up with Locales/en.

can be run by `npm run check-locale -- [option]` where option specifies which locale to work on.
The script will take the last update done on the specified locale directory, and runs git diff on Locale/en from that commit hash to the current HEAD,  which should produce all changes made to Locale/en since last time the specified locale directory has been updated.

Pros: simple implementation with precise diff without adding more dev dependency in package.json
Cons: Only gives diff from last commit on working locale, so if last commit has not completely caught up with en, those missed changes will not appear in the diff.


mizma:
I may need to re-write the shell script with .ts instead so it works on Windows. but I would like to know if this even is a good idea.

mizma:
Todo:

- [ ] Rewrite in javascript
- [ ] add option to specify commit hash
- [ ] add option to specify number of commit history to go back against specified locale
- [ ] also output commit log of the specified commit hash (so user can know what commit it is diffing against.

mizma:
I think it is ready, but will wait till 0.7.11 is released.

arntsonl (approved):
This is great thank you! Also gives us the ability to add more scripts if we ever need them :)

PR #1339, opened 2025-04-05
Web Config Saves Fix  + Removing Legacy Code

arntsonl:
Moved config mode check to the DriverManager

Removed ConfigManager as this idiom never took off and we have better ones in place

Modified NetDriver to run rndis making it the controller of web

Moved Storage -> config Mode to Driver -> config Mode

Fixed no-save issue with web config calls


arntsonl (commented):
Looks correct, going to merge so we can narrow down the crash

PR #1340, opened 2025-04-05
Adds an advanced tab where individual macro definitions as json can be copied/import

Pelsin:
Also added some validations if one where to import more than maximum inputs / incorrect types

https://github.com/user-attachments/assets/725cb53b-7ee5-46ce-bdff-fffc70e81e43


arntsonl (approved):
Really cool little feature, doesn't change any of our underlying structures and I think this is a pretty easy one to throw in.

PR #1342, opened 2025-04-07
Updated Tilt addon to be profile friendly

mikepparks:
Added event handler to Tilt to reload inputs when profile is changed.
Fixed issue with Tilt input mapping not correctly resetting the map on event.


arntsonl (approved):
Good use of the events!

PR #1343, opened 2025-04-07
Web Server RNDIS Fix

arntsonl:
Fix for RNDIS memory leak (web server crash)


arntsonl (commented):
I added this in when I did the case rgb leds, my fault!

PR #1345, opened 2025-04-07
Fixed PS4 mode support for Gamepad Host

mikepparks:
Changed magic numbers for PS4 device IDs to use defines in Gamepad Host.
Added main PS4 "Console" IDs to Gamepad Host device check to allow both modes to function.


arntsonl (approved):
Perfect, yeah good we catch both of these

PR #1347, opened 2025-04-08
Update Ultra Lite and Zen configs

TheTrainGoes:
Adjusts LEDs on Ultra and Lite down to 200 from 255 due to brightness fix in RC2.
Lowers number of brightness steps from 10 back down to 5.
Changes the Zen aux buttons to squares like on the real controller.


arntsonl (approved):
Looks good, much better max light for these controllers

PR #1350, opened 2025-04-09
Standalone D-Pad/Analog updates

mikepparks:
Fixed handling of digital only dpad inputs to correctly separate them while allowing them to affect lighting and display.

Updated labels for standalone analog and tilt modifiers in webconfig for more consistency.


arntsonl (approved):
Looks good!!

PR #1352, opened 2025-04-09
Disable Tilt if no pins mapped.

mikepparks:
Disable Tilt functionality when addon is either disabled dynamically or pin mappings for percentage modifiers are unset.


arntsonl (approved):
Looks clean, nice easy fix

PR #1365, opened 2025-04-14
Switch Pro Controller

mikepparks:
Very basic implementation of the Switch Pro controller.  Most advanced features (IR/NFC) are not functional, and motion is a work in progress. Implementation based on research by [dekuNukem](https://github.com/dekuNukem/Nintendo_Switch_Reverse_Engineering) and [Chromium gamepad handling source](https://chromium.googlesource.com/chromium/src/+/b37eb129399993dd308eddf033509a03cc540e62/device/gamepad/nintendo_controller.cc).  Erratic behavior in joy.cpl and other gamepad testers is known and matches behavior from a true Pro Controller as other testers do not fully implement/ignore the reports.


arntsonl (approved):
Looks fantastic, adds not only switch pro but the aux power management as well which we can start adding to other

PR #1371, opened 2025-04-24
Input Mode Subtypes

mikepparks:
Adding subtype/special controller support to Input Modes when available. This will allow controllers to simulate other non-gamepad/stick controller types if the mode allows for it.

Controller types currently planned:
* Wheel
* HOTAS/Flight
* Guitar
* Drum

The only input subtypes that have been implemented in this PR are those that are derivatives of the existing gamepad controller for the platform.  This means that the base HID input report structure is the foundation of other subtypes.  There are some edge cases where additional modifications need to be made on a mode by mode bases, such as the vendor/product ID for specifically selecting a PS4 wheel.


arntsonl (approved):
Looks great! Some good QoL on the web config part, lots of new options with our ps3/ps4 subtypes. Once we get ps5 rolling, we'll have tons of stuff :)

arntsonl (approved):

PR #1378, opened 2025-05-09
Focus mode toggle hotkey

eliasreid:
Added a hotkey action that allows for forcing on "focus mode". This lets users without access to a physical switch take advantage of the focus mode feature via button combo.

Maybe there's a better name for the `toggleState` field of `FocusModeOptions`. Maybe this should be called something like `forceActive` or `focusModeOverride` instead?


arntsonl (approved):
That's a clever feature, I like it! Putting it in hotkey form like that and only requiring 1 button makes it much easier to have a focus button or hotkey combo

PR #1385, opened 2025-05-19
Add additional board defined layout slot

TheTrainGoes:
This PR adds another board defined layout slot for boardconfig.h.  This also adds a layout for the Haute42|COSMOX E16 Southpaw.


mikepparks:
Looks good here.

mikepparks (commented):
One more to the pile.

PR #1394, opened 2025-06-02
Update cmake.yml

Fortinbra:


arntsonl (approved):
Updating the v7 script which updates our builds

PR #1397, opened 2025-06-10
Update profile to 6

TheTrainGoes:


src/config_utils.cpp
```diff
@@ -1335,7 +1335,7 @@ void gpioMappingsMigrationProfiles(Config& config)
         }
     };
 
-    for (uint8_t profileNum = 0; profileNum <= 2; profileNum++) {
+    for (uint8_t profileNum = 0; profileNum <= 4; profileNum++) {
```
bsstephan, line 1338: One thing you might want to do is add a
`const static uint8_t MAX_PROFILES = 6` to `Storage` in `storagemanager.h` and then replace these 4s with `Storage::MAX_PROFILES-2` just to remove the magic number from the code when it's read.

lib/nanopb/extra/requirements.txt
```diff
@@ -1,2 +1,2 @@
 protobuf>=3.19
-grpcio-tools==1.71.0
+grpcio-tools==1.71.0
```
bsstephan, line 2: I think you can probably just drop this commit from your branch to get rid of this change.

bsstephan (approved):
A couple "nice to have"s but I think this otherwise looks perfect and is ready to merge whenever.

PR #1401, opened 2025-06-11
separate all analog settings per stick

### LINKED ISSUE #1244: Make deadzones (or all analog settings) configurable per joystick
### Problem

Currently settings like deadzones or auto calibration have to be the same on both joysticks, which is problematic when two mismatched joysticks are used (in my case  i am using one linear slider and one joy-con-like joystick which requires ~65% outer deadzones)

### Solution

make those settings separate on both joysticks

i could try

NickGuyver:
Make all the analog settings assignable per stick, tabulate the analog webconfig settings, and clean up the formatting. Closes #1244 .


arntsonl (dismissed):
Yeah I really like this, makes it much easier to separate out the logic of each analog stick.

arntsonl (approved):

Issue #1381, opened 2025-05-17
Web Config: Unchecking XInput -> Advanced USB Override Leaves  "Override Physical IDs" checked


Paxtez:
### Expected Behavior

Unchecking a value that hides another value should turn it off also.

### Current Behavior

If you uncheck  "Advanced USB Override" while "Override Physical IDs" is checked, the "Override" section gets hidden but stays selected, even when saving.

And leaves it as a franken-controller:

```
      ========================== Summary =========================
Vendor ID                : 0x10C4 (Silicon Laboratories, Inc.)
Product ID               : 0x82C0
Manufacturer String      : "©Microsoft Corporation"
Product String           : "Controller"
Serial                   : "00706FA"
...
        +++++++++++++++++ Device Information ++++++++++++++++++
Device Path 1            : \\?\USB#VID_045E&PID_028E#00706FA#{a5dcbf10-6530-11d2-901f-00c04fb951ed} (GUID_DEVINTERFACE_USB_DEVICE)
```

### Context
Version: Pico (GP2040-CE_0.7.11_Pico.uf2)
Current: v0.7.11
Window Chrome

### Steps to Reproduce
1. Web-config -> Settings -> Input Mode Settings.
2. Current Input Mode -> Xinput -> Check "Advanced USB Override"
3. Check "Override Physical IDs"
4. Save -> Reboot -> Web-config
5. Settings -> Input Mode Settings
6. Uncheck "Advanced USB Override"
7. Save -> Reboot -> Controller

### Possible Fix
I've yet to build the project, and I'm not familar with JSX, but looking at the source this might be a possible fix.

www/src/Pages/SettingsPage.jsx
After Line 1021:
Add:
```
		if (!e.target.checked) {
			setFieldValue('usbOverrideID', 0);
		}
```

But perhaps a more elegant solution might be to check during save or even load, or to disable the top checkbox if the bottom one is enabled.

PR #1406, opened 2025-06-17
fix issue 1381

NickGuyver:
Fix issue [1381](https://github.com/OpenStickCommunity/GP2040-CE/issues/1381)


arntsonl:
Nice, easy fix!

PR #1418, opened 2025-07-07
Add option to select what config to build from github actions

Pelsin:
Screenshot 2025-07-07 at 02 07 44

When selecting a configs it should skip the others, should drastically reduce the build time when only building one config


Fortinbra (approved):
Amazing!

PR #1420, opened 2025-07-08
Fix for rndis crashing the i2c

arntsonl:
Moving rndis init slightly and removing tusb_inint from the rndis driver and lib

This was an issue with tinyusb ports being wrong when rndis_init would call tusb_init(); (thank you Peter Harper!)

### DIFF
--- a/lib/rndis/rndis.c
+++ b/lib/rndis/rndis.c
@@ -213,8 +213,8 @@ void tud_network_init_cb(void)
 
 int rndis_init(void)
 {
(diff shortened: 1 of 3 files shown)


arntsonl:
This was quite an old bug in our code. rndis_init would call tusb_init() and run tinyusb tasks before we had a chance to do anything.

SDK 1.5.1 would run tinyusb specifics after we had a chance to finish the core code, SDK 2.1.1 was running tinyusb before we had a chance to reconfigure TinyUSB.

PR #1421, opened 2025-07-10
code USBHID button display w/generic numbers rather than PS buttons

bsstephan:
I realized a while back that I never did this and finally got back around to it. I feel like the button numbers are more useful for USBHID users than the old PS3-style PS coded buttons.

### DIFF
--- a/headers/display/ui/screens/ButtonLayoutScreen.h
+++ b/headers/display/ui/screens/ButtonLayoutScreen.h
@@ -15,11 +15,11 @@
 #include "GPGFX_UI_layouts.h"
 
 #define INPUT_HISTORY_MAX_INPUTS 22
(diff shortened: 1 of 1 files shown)


bsstephan:
Making this a draft because I missed that it broke PS3 output (they used the same index, I need to fix that) but also apparently somehow affected PS4/5 output? Need to investigate further.

bsstephan:
Fixed both issues, an entry for the PS5 input mode was never made, so it was defaulting to index 0, which was the entry originally for PS3+Generic, that I changed without thinking, giving both PS3 and PS5 the Generic button numbers. So, I made a new entry for Generic and pointed PS5 at PS4.

arntsonl:
Looks good to me, ps5 mode going to generic bug makes sense.

arntsonl (approved):

Issue #1386, opened 2025-05-19
Saving USB Host settings with GP28=D-, GP29=D+ freezes RP2040-ZERO board


kiyokura:
### Expected Behavior
The USB host port should function correctly when GP28 is wired to D- and GP29 to D+.

### Current Behavior
I wired the board as shown below and configured the settings via the web interface.

- Wiring
  - D+ : GP29
  - D- : GP28
- Web Config
  - Pin: 29
  - Order: 1 (D-/D+)

Clicking the Save button after configuring the settings via the web interface causes the board to freeze and the interface to become unresponsive.

I also tried disconnecting and reconnecting the USB cable, but the GP2040-CE firmware no longer booted.
To recover the board, I had to run flash_nuke and then reflash the GP2040-CE firmware.

Note: In all of the following configurations, the system behaved as expected.
(Successfully authenticated on PS4)

- Pattern 1
  - Wiring
    - D+ : GP28
    - D- : GP29
  - Web Config
    - Pin: 28
    - Order: 0 (D+/D-)
- Pattern 2
  - Wiring
    - D+ : GP27
    - D- : GP28
  - Web Config
    - Pin: 27
    - Order: 0 (D+/D-)
- Pattern 3
  - Wiring
    - D+ : GP28
    - D- : GP27
  - Web Config
    - Pin: 28
    - Order: 1 (D-/D+)

### Context

* Version : Waveshare Zero (GP2040-CE_0.7.11_WaveshareZero.uf2)
* Board : RP2040-ZERO (Clone)
* OS : Windows 11
* Browser : Chrome

### Steps to Reproduce

1. Wiring : GP29 -> D+ and GP28 -> D-
2. Web Config : Pin -> 29 and Pin Order -> 1(D-/D+)
3. click `save` button

speedypotato:
Also encountered with GPIO 3, and -1 order on my boards.  I have saved and rebooted to webconfig and seen failed enumeration, and replugging shows that the flash got reset.  I have also seen it successfully save and reboot to webconfig, but then booting again to controller mode fails enumeration and requires a flash nuke or force webconfig.

https://github.com/OpenStickCommunity/GP2040-CE/pull/1424 should hopefully resolve the issue.

IceyMu:
I'm also having the exact same behaviour with the same version of the firmware which is preventing me from authenticating on PS5. I've been able to work around this by downgrading to version 0.7.8 (I haven't tried the in between versions)

[...]

PR #1424, opened 2025-07-13
pico_pio_usb to 0.7.2 and changed its and neopixel's pio and sm assignments

speedypotato:
Should fix https://github.com/OpenStickCommunity/GP2040-CE/issues/1386 and https://discord.com/channels/1049366310389289001/1392043096526684252 .

Essentially if USB Host is touchy, let's not try to change the default PIO and SM it's on as this is likely the most tested use case.  We should move the neopixel pio instead.  I've also updated pico_pio_usb to the latest 0.7.2 release and confirmed auth with a Haute42 Booter5 on https://www.jfedor.org/ps-auth-tester/


arntsonl:
Mentioned it in Discord, but unfortunately this doesn't work on the Magicboots PS4 v1.1 :(

arntsonl:
Converting to a draft because we know magicboots are still an issue :(

arntsonl:
This is a deeper issue in TinyUSB :( Hathach is trying to resolve it in the issue here: https://github.com/hathach/tinyusb/issues/3149 but we're still having mega issues

PR #1429, opened 2025-07-20
Themes default check on startup

### LINKED ISSUE #1428: RGB LED Static Theme not being saved on reboot
### Expected Behavior

When RGB LED Static Theme is selected on the controller, the same Static Theme selected should display on reboot.

### Current Behavior

After changing the LED Animation to Static Theme and choosing any different Static Theme, on reboot, the controller defaults to the `themeGuiltyGearTypeA` static theme. When this occurs, the other LED settings (i.e. brightness and pressed color) are saved. This occurs in all input modes.

### Context

* Firmware Version: 0.7.11
* Board Config Name: GP2040-CE_0.7.11_Haute42COSMOX.uf2
* Device Name: Haute42 Hautepad S16

### Steps to Reproduce

1. Plug in controller, booting into any input mode
2. Change LED Animation to Static Theme
3. Change Static Theme to any option other than `themeGuiltyGearTypeA`
4. Unplug and plug controller back in
5. LED theme will be Static Animation - themeGuiltyGearTypeA

Pelsin:
Should fix #1428


arntsonl (approved):
Nice, theme index was in the wrong place it looks like

PR #1432, opened 2025-07-22
Update web-config labels for dedicated analog inputs

TheTrainGoes:
This is a very small PR that updates the labels for the dedicated analog inputs to include their overall direction as well.


arntsonl (approved):
Makes sense!

PR #1443, opened 2025-08-03
Add status bar warning when no USB data detected

TheTrainGoes:
This PR targets devices with an OLED screen but no onboard LED set to flash when a USB cable is connected without data lines or when there is a potential issue with the USB port.  

When USD data cannot be established this will replace the status bar on boot to read `WARNING: NO USB DATA`


TheTrainGoes:
Closing this as it was better done in https://github.com/OpenStickCommunity/GP2040-CE/pull/1460

PR #1445, opened 2025-08-06
Add PCEngine/TurboGrafx-16 Addon

lpalombo:
Adds an addon to allow PC Engine/TurboGrafx-16 controllers to be used as inputs in GP2040-CE. This works functionally the same as the SNES addon. It's also loosely based on the [Raphael Assenat's TurboGrafx-16/PC-Engine/Duo to USB adapter source code](https://www.raphnet.net/electronique/tgfx16_usb/index_en.php).

As a proof-of-concept, I soldered up a 8-pin mini DIN connector using [this guide](https://github.com/darthcloud/BlueRetro/wiki/PCE-&-TG16-interface) and tested out some PC Engine controllers. Works great!

I made sure to add a web-configuration component to the addons page and english copy, let me know if I missed anything!

https://github.com/user-attachments/assets/bbf42413-17ec-48ed-849e-6e9e1030858a


mikepparks:
I see you're testing on an Avenue 3. Have you confirmed that this works with a 6 button pad?

lpalombo:
I have not! Nor do I have access to one, but I'll look around and double check my source to see if it'll be supported anyways.

lpalombo:
I just made a commit that adds support for 6 button controllers. Not sure exactly if it works but I do have a connection with a controller so I'll test it out ASAP and report back. 2 button controllers still work fine though

PR #1450, opened 2025-08-09
Update cmake.yml

Fortinbra:


arntsonl:
I believe we had this in a few other PRs, so we should be good to close.

PR #1461, opened 2025-08-19
Add layout for C16 + C16-S and fix issue with board defined layouts

TheTrainGoes:
This PR adds a layout for the Haute42|COSMOX C16-S.

This PR also fixes an issue where, if you only had one board defined additional layout, it would not show the right-side option in web-config.


TheTrainGoes:
Also added a layout for the C16 which has two buttons swapped from the M Ultra.

arntsonl (approved):
Looks good! I'm going to update C16-S to pills once this is in.

TheTrainGoes:
@recklessXRP you were answered on the Haute42|COSMOX Discord, please do not post the same thing in multiple places.

PR #1480, opened 2025-09-03
Disable USB error event until better detection method is written

mikepparks:


arntsonl (approved):
We'll wait on error messaging until we got this locked down a bit more.

PR #1498, opened 2025-09-18
Fix swapped LS and M1 buttons in C16 layout

Fadel2000gazal:
Fixed an issue where LS (Pin 18) and M1 (Pin 26) were swapped in the layout, causing the layout screen to appear incorrectly mapped.


TheTrainGoes:
@Fadel2000gazal ,

This is not an issue.  This is the correct layout for the Haute42|COSMOX M Ultra.

The C series units have a different button location for those two.  I have already updated with unique layouts for this.

Fadel2000gazal:
@TheTrainGoes, 

Sorry about that, and thanks for clarifying!

So, would it make sense to create a new config specifically for the C16 and C16-S, so it won’t cause any conflicts with the Haute42|COSMOX M Ultra layout?

TheTrainGoes:
@Fadel2000gazal 

The corrected layouts for the C16 and C16-S are now in the board defined alternate layouts under the M Ultra config.  

At this time I'm not interested in doing more individual Haute42 configs.

PR #1505, opened 2025-09-23
Update ko-KR localization strings

losernator:
* Update Korean strings for next release


mizma:
Have you checked the en locale change for xinput-mode-text in  www/src/Locales/en/SettingsPage.jsx?  seems the wording has changed since last commit (checking using `npm run check-locale -- -l ko -f 1` or `node ./scripts/checklocale.js -l ko -f 1` on Win)

losernator:
> Have you checked the en locale change for xinput-mode-text in www/src/Locales/en/SettingsPage.jsx? seems the wording has changed since last commit (checking using `npm run check-locale -- -l ko -f 1` or `node ./scripts/checklocale.js -l ko -f 1` on Win)

I just change some phrase. thanks!

PR #1508, opened 2025-09-25
Update zh-CN localization

rebornsick:
- Updated Chinese translation for all new added strings
- Added translation for new addons
- Unified the translation style


arntsonl (approved):
Looks great!!

PR #1509, opened 2025-09-27
Turkish Localization

NogayHan06:
This is my first time uploading localization for another project. If I've uploaded it incorrectly, could you please explain how to do it?

Also, if appropriate, when will you add localization to the project? Thank you.


mizma:
Great addition!

A few minor things I noticed.

- It's advised to not have untranslated strings in your localization file.  If it should stay in English, it is better to omit the item from the localization file so it can automatically fall back to the English version and stay up-to-date with any changes in English (Things like Unique console names that stay in English in the locale, button names that does not require translation in your locale, or strings you are not ready to translate for the time being etc.)
- It might help to add all the translation for 'language-selector' items in other languages.  Especially for en locale which is probably mandatory for all other languages to fall back to.

You could also check the merged PRs with initial language addition (e.g. https://github.com/OpenStickCommunity/GP2040-CE/pull/1048)

As for when this will be pulled, I think will be up to the project owners.

NogayHan06 (commented):
Dear @mizma, I've added the fixes you mentioned to my localization project. Thank you for your help. Are there any other issues you see?

mizma:
Hi, @NogayHan06 
The language names in the Components.jsx I think has the format of 
"The language name in it's own language (language name in current locale)"
so, for Turkish I think it can be 
Türkçe (Turkish) for English translation, Türkçe (トルコ語) for Japanese translation, while in the Turkish locale file, just "Türkçe" should be enough... I think (the last part is currently not consistant across all languages).
For people trying to switch to a language, I think we want the language in it's native form be listed so someone who only understand that one language can locate his/her language while the part in parenthesis gives a bit of context in the current language.  At least that is how I understood this section.

NogayHan06 (commented):
I misunderstood the part you mentioned :) I edited it based on your example. Thank you very much for your help. @mizma

NogayHan06 (commented):
tr-TR localization, some incorrect words corrected..

arntsonl:
Waiting for after 0.7.12 for this one, will look at it when we get RC done

arntsonl (approved):
Looks good! we'll need to update check locale at some point for new languages, but its fine for now.

PR #1522, opened 2025-10-20
20251020 Remove DuelPad Zen config

TheTrainGoes:
This PR removes the DualPad Zen config.

DuelPad customers have been unhappy with our implementation of the mini-menu compared to the fork that DuelPad maintains.  Our release has also caused confusion and upgrade issues for their customers so we are removing it.


arntsonl:
Totally agreed! We're not supporting their commercial product.

PR #1523, opened 2025-10-20
20251020 Update to allow board defined layouts to have names in web-c…

TheTrainGoes:
…onfig


www/src/Pages/DisplayConfig.jsx
arntsonl: Yeah the i8n check stuff actually masks our missing translations if they are not defined, so we don't want to do this. But you can just revert this file's change.

Issue #1231, opened 2024-12-12
Tilt addon Pin Settings not changing across different Pin Mapping Profiles 


jack2game:
Pulled from the latest main branch.

### Current Behavior

Despite you can have different pin settings for Tilt addon across multiple profiles, only Profile 1 will be valid, and the settings in Profile 1 will still be working even when you are in a different profile.

### Steps to Reproduce

Try different pin settings as per screenshot below for Profile 1 and Profile 2. When in Profile 2, the Pin Settings in 1 will be working.

### Screenshots & Files

![chrome_1734019208_EALOrBSRK1](https://github.com/user-attachments/assets/44d8cc50-d79e-4edb-ab42-ed4202b25612)

![chrome_1734019214_QK7hpqV3pd](https://github.com/user-attachments/assets/cdf1605f-2247-4e24-99f9-7b35d9a2e86b)

jack2game:
@mikepparks 

Unfortunately the recent fix actually broke the addon when switching profiles.

https://github.com/OpenStickCommunity/GP2040-CE/blob/84ab80b01f84bc0a409bcb449d4291a220dd74b6/src/addons/tilt.cpp#L96-L97

The 2 lines here will remove the pinmapping, however it sometimes cannot be recovered when I switch back to the profile.

My setup:
- Hotkey of `Fn` + `DPad Up` = Switch to next profile
- Profile 1 has tilt keys
- Profile 2 doesn't

When I switch to profile 2 and then back to profile 1, it often completely lose the Tilt functions.

Issue #1507, opened 2025-09-23
Menu Select opening menu does not respect profile changes


Yarn:
### Expected Behavior

Menu Select will open the menu when the button corresponding to menu select in the current profile is pressed.

### Current Behavior

Menu Select will open the menu when the button corresponding to menu select in the profile selected at boot is pressed.

### Context

Tested on 7.11 and 7.12 beta 2 on a Haute42 T16
Haute42 COSMOX

### Steps to Reproduce

* Add a second profile and a way to switch between profiles.
* Add a Menu Select button to one of the profiles.
* Boot into the profile with menu select
* switch profiles
* press the key menu select was bound to.

### Screenshots & Files

[menuselectprofileissue.gp2040.json](https://github.com/user-attachments/files/22502863/menuselectprofileissue.gp2040.json)

TheTrainGoes:
I have tested and validated this bug with the provided backup but find the pinout and options very difficult to navigate.

@Yarn is there a specific reason for the layouts in both of these profiles or is it just to highlight the bug?

Yarn:
The layout is just what I'd set up that made sense on my controller. I don't think it's relevant to the bug. I'm pretty sure any setup where you can switch profiles and have menu select that isn't the same button on all profiles will reproduce the bug.

TheTrainGoes:
I've tested this PR and everything seems to working as intended!

CapnHawke:
Just tested this issue on the new PR and I'm confirming that it works as intended in the PR.

PR #1525, opened 2025-10-22
Refactor gamepad profile handling and update display mappings

Pelsin:
Should fix: https://github.com/OpenStickCommunity/GP2040-CE/issues/1231 https://github.com/OpenStickCommunity/GP2040-CE/issues/1507

### DIFF
--- a/headers/addons/display.h
+++ b/headers/addons/display.h
@@ -197,17 +197,13 @@ public:
     virtual void reinit() {}
     virtual std::string name() { return DisplayName; }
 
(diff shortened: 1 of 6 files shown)


CapnHawke:
Tested on RP2040 advanced breakout board and this is working great. 

Thank you for handling this.

arntsonl (approved):
Nice, that's a great way to fix it! Fixes the logic with our booleans. Now its just a simple "last profile != current profile" reinit!

PR #1527, opened 2025-10-23
feat: Add manual calibration feature and HML hardware support

leonxis:
Features:
To prevent potential errors from automatic calibration, a manual calibration option is provided. Once manual calibration is performed, when automatic calibration is disabled, the manually calibrated center point will be automatically applied.

- Add independent manual calibration for analog sticks 1 and 2
- Implement 4-step calibration process (top-left, top-right, bottom-left, bottom-right)
- Add complete i18n support (English/Chinese) for calibration messages
- Add new HML hardware configuration with custom GPIO mapping
- Support parameterized messages for dynamic content

Technical changes:
- Add joystick_center_x/y/x2/y2 fields to config proto
- Add manual calibration API endpoints (/api/getJoystickCenter, /api/getJoystickCenter2)
- Update analog input logic to support manual calibration values
- Add comprehensive frontend calibration interface with step-by-step guidance
- Create HML hardware configuration with custom pin mapping and analog stick setup

Next step plan: 
- provide a manual calibration feature in the mini LED.


TheTrainGoes:
Hi @leonxis ,

Thank you for the PR submission.  

What is the HML hardware?  Is it a custom device?  Can you please link to the repo for it so we can take a look?

mikepparks:
I would ask that you recreate this PR after removing the HML config.  Also, any code commentary should be written in plain English, including the TSX files. The defines also do not appear to be in use for default values, so it feels like this needs some extra work before we can look at it.

PR #1529, opened 2025-10-24
Fixes dpad to analog being interrupted by mouse movement input to resolve #1515

### LINKED ISSUE #1515: Keyboard Add-on: Mouse Inputs interrupt keyboard input.
> Please read these quoted sections before submitting, and remove them (including this one) when you have provided what
> is asked, in order to make addressing the issue easier for everyone. Thank you!
>
> Be sure to:
> * Confirm the feature does not exist in the latest version of the firmware, found
  [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases).

The problem I am describing is only on the compiled version of the master branch, this report is only to notify of a bug in an upcoming release (I pressume) relating to using mouse input as a joystick input.

### Problem

> Which problem is this feature request solving?
>
> Provide a clear and concise description of what the problem is. For example, "I'm always frustrated when..."

When using Left Joystick (mapped to WASD) and Right Joystick (mapped to Mouse Input). Mouse input interrupts WASD joystick as a complete reset to 0,0. Requires WASD re-pressing to resume untill interrupted again.

### Solution

> Describe the solution you would like to see
>
> Provide a clear and concise description of what you want to happen. Let us know about other solutions you've tried or
> researched.

In order to use Mouse/Keyboard on as an example, Splatoon on Switch, the right joystick (mouse input) must not interrupt and reset the WASD keypress.

> Beyond a simple yes or no, we are also willing to help with partial solutions or to refine a proof of concept, if you
> are able to contribute.

I will try to triage and find why this behavior is happening, but I do not know how much time I can devote to it. Otherwise, this feature would allow mouse/keyboard on many controler only FPS-like games.

Fran89:
This is my attempt to fix the problem of the dpad interruption whenever the mouse report movement input. The issue lies in the preprocess_report function. Inside it uses a function to zero out the dpad input, no doubt useful when keyboard input is pressed, however the mouse_report() also uses it when updating right analog input and thus both dpad to left analog and mouse right analog cannot be used at the same time (extremely useful for mapping first-person shooters, to gyro controls on Switch). I have included Pelsin's updated code and moved the dpad zero function to the line after the preprocess call on kbd_report() and it seems to have fixed the problem.

However, I believe this need testing as I'm not sure of potential conflicts, or edge-cases that might arise from this change in code. I'm only pull requesting it so that others might test it, or help in fixing this. Currently, I have tested it very limitedly on both PC and Switch.

### DIFF
--- a/src/addons/keyboard_host_listener.cpp
+++ b/src/addons/keyboard_host_listener.cpp
@@ -174,13 +174,20 @@ uint8_t KeyboardHostListener::getKeycodeFromModifier(uint8_t modifier) {
 
 void KeyboardHostListener::preprocess_report()
 {
-
-  _keyboard_host_state.dpad = 0;
   _keyboard_host_state.buttons = 0;
-  _keyboard_host_state.lx = joystickMid;
-  _keyboard_host_state.ly = joystickMid;
-  _keyboard_host_state.rx = joystickMid;
-  _keyboard_host_state.ry = joystickMid;
+  // --- preprocess only select analog movements --- // (by Pelsin)
+  if (mouseMovementMode == MOUSE_MOVEMENT_LEFT_ANALOG) {
+    _keyboard_host_state.lx = joystickMid;
+    _keyboard_host_state.ly = joystickMid;
+  } else if (mouseMovementMode == MOUSE_MOVEMENT_RIGHT_ANALOG) {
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
This makes sense to me

PR #1537, opened 2025-11-03
Move board button layout options and update display configuration formatting

Pelsin:


Pelsin:
I'll close this as i dont think it's relevant any more

PR #1546, opened 2025-11-09
Add P5General mode

besavior-team:
Added P5General mode:
* Currently supports all PS5 games, but without rumble
* Requires connecting the P5General dongle to USB host port
* If you have a P5General dongle, please upgrade its firmware to [p5g.25110916.zip](https://github.com/user-attachments/files/23439800/p5g.25110916.zip) before testing, as this will provide a better polling rate
* I2C0 needs to be disabled; otherwise, the report rate will be extremely low. The cause has not yet been analyzed

Thanks


TheTrainGoes:
Hi @besavior-team ,

Thank you for the updated PR.  I have tested this and it has addressed a few of my previous concerns.  

I will test this out some more once we get v0.7.12 out.

besavior-team:
> This is fantastic work, thank you @besavior-team !!! Coding flows exactly how we do it, naming is good with DualSense (General), looks good.

Thank you very much, and have fun everyone!!!

PR #1550, opened 2025-11-13
Add new languages to checklocale

mizma:
Added following locales
- tr-TR
- fr-FR


mizma:
will wait for other translations in the pipeline

arntsonl (approved):

PR #1567, opened 2025-12-19
Update PXPGamepad BoardConfig.h

MegaBitmap:
This fixes the default config for the second analog stick.
This is needed because of this change: #1401


arntsonl (approved):
Nice! This should hopefully play well with our new analog calibration stuff

PR #1574, opened 2025-12-29
add display settings for osumgp-rp2040

NickGuyver:
Add display settings for osumgp-rp2040 for use with SparkFun LCD-22495 which is non-standard size


mikepparks:
This looks great.  Good catch on adding the defines for toggling status bar items.

mikepparks (approved):

PR #1590, opened 2026-01-26
screen saver fixes

NickGuyver:
have screen saver exit match entry

### DIFF
--- a/src/display/ui/screens/DisplaySaverScreen.cpp
+++ b/src/display/ui/screens/DisplaySaverScreen.cpp
@@ -52,14 +52,11 @@ void DisplaySaverScreen::drawScreen() {
 
 int8_t DisplaySaverScreen::update() {
     if (!DriverManager::getInstance().isConfigMode()) {
-        uint16_t buttonState = getGamepad()->state.buttons;
-        if (prevButtonState && !buttonState) {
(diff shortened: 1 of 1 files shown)


src/display/ui/screens/DisplaySaverScreen.cpp
```diff
@@ -52,14 +52,11 @@ void DisplaySaverScreen::drawScreen() {
...
     if (!DriverManager::getInstance().isConfigMode()) {
-        uint16_t buttonState = getGamepad()->state.buttons;
-        if (prevButtonState && !buttonState) {
-            if (prevButtonState != 0) {
-                prevButtonState = 0;
-                return DisplayMode::BUTTONS;
-            }
+        Gamepad * gamepad = Storage::getInstance().GetGamepad();
+        
+        if (gamepad->state.buttons || gamepad->state.dpad) {
```
arntsonl: hmmm is it okay we got rid of prevButtonState? do we need to add the dpad as part of our previous state? I'll need to test this one

arntsonl (commented):
I'll need to test this one, odd bug but that would be a nice easy fix. We could also potentially get rid of prevButtonState maybe?

arntsonl (approved):
Makes good sense to me!

PR #1591, opened 2026-01-27
add brightness control

### LINKED ISSUE #1536: Settings for dimming the screen/OLED display
Since many controllers are using an OLED display which is prone to burn-in it would be great if it was possible to dim the screen to reduce the burn-in rate. It would also be better for playing in a dark room since the display can get a bit bright.

Maybe a display dimming setting in the web config, and possibility to map buttons or shortcuts to screen brightness up/down, and even a screen toggle off/on, as well as a setting in the mini display menu.

I can contribute by testing firmware on my Haute42 Ultra M gen 2 device.

NickGuyver:
add display brightness control via webconfig, resolves issue 1536


arntsonl:
There's an odd bug on first install / start. When you flash this onto a Haute and go into the web config, then click "Save" without clicking the brightness, it gives you this.

image

arntsonl:
I just put this onto an upgraded Haute, and I get the following:

image

arntsonl:
Also moving brightness down below about 15% makes the screen unreadable. So maybe limit the lowest percentage to like 20%?

arntsonl (approved):
Looks great, thank you @NickGuyver ! There's a bug/issue that Pelsin and I talked about with preview display options update, we'll fix that in another PR. Nothing related to your changes, thank you for fixing this up!

PR #1592, opened 2026-01-28
bump sys clock to max supported

NickGuyver:
verified working
- usb keyboard host
- xbox dongle auth
- oled display


arntsonl:
I'll need to do a ton more testing but this is neat!

arntsonl (approved):
Yes! This is an awesome SDK feature.

PR #1593, opened 2026-01-29
issue 1373

### LINKED ISSUE #1373: Can supports hotkey assignment for SNES pad buttons?
### Problem

Using SNES Extension, the buttons on the SNES Pad cannot be used as hotkeys at v0.7.11.

### Solution

I'd love to be able to assign hotkeys to the buttons on the SNES pad.

No I don't...

NickGuyver:
fix for issue 1373


arntsonl (approved):
That actually makes quite a bit of sense. Hotkeys can come from add-ons and not the other way around, a hotkey cannot make an add-on input happen.

PR #1595, opened 2026-01-31
Add display layouts for Haute42|COSMOX X-Move

TheTrainGoes:
This PR adds three new board defined display layouts to the Haute42|COSMOX M Ultra config.

Those layouts are used for the new Haute42|COSMOX X-Move.  It will allow for all current combinations of modules for the X-Move:

1 - Stock config
2 - WASD module installed
3 - Joystick module installed


arntsonl (approved):
Looks good!

PR #1601, opened 2026-02-05
fixes for 1496 & 1503, webconfig cleaning

### LINKED ISSUE #1496: ADS1256 addon web configuration
I think there may be a bug in the add-on config: Analog Max always stays at 3.3v. Changing to 5V does not seem to work.

1 select 5V
2 save
3 refresh page
result: Analog Max is still 3.3v in the web config

Image

Also, as pictured above GP18 is DRDY, yet it does not show as assigned to addon, plus it always shows as triggered:

Image

Version
Pico (GP2040-CE_0.7.11_Pico.uf2)
Current: v0.7.11
Latest: v0.7.11
Board: RP2040

### LINKED ISSUE #1503: Can't save Rotary Encoders configuration / Can't disable Encoder 1/2 if i Enable them !
Can't save Rotary Encoders configuration / Can't disable (Encoder 1/2) (Allow Wrap-Around) if i Enable them !

- ULTIMATE PICO RP2040 128 MBITS 16 MO
- KY-040 Module encodeur rotatif

Image

Image

Image

NickGuyver:
fixes #1496
fixes #1503

and then some webconfig cleaning


headers/addons/spi_analog_ads1256.h
```diff
@@ -49,7 +49,7 @@ class SPIAnalog1256Input : public GPAddon {
     float values[ADS1256_CHANNEL_COUNT]; // Cache for latest read values
     bool enableTriggers;
     uint8_t readChannelCount; // Number of channels to read from the ADC
-    float analogMax = ADS1256_MAX_3V;
+    float analogMax = ADS1256_MAX_3V * 10;
```
arntsonl: this should be * 10.f to make sure you don't cast it to an int on multiply

arntsonl (changes requested):
Two changes and its ready!

arntsonl (approved):
Fantastic!

PR #1608, opened 2026-02-17
sinput mode initial commit

### LINKED ISSUE #1438: SInput HID Protocol Support
### Problem

> Which problem is this feature request solving?
>
Currently: the firmware covers various Controller Types, but each Controller Type will function differently and may be missing features like Motion Sensors or losing out on Analog Triggers, or additional button inputs. HID USB is very close to a more modular setup.

### Solution

> Describe the solution you would like to see
>
Not to long ago: [Hand Held Legends](https://handheldlegend.com/), the folks behind the upcoming GC Ultimate, introduce a new H[ID Protocol specifically designed for SDL-based Apps and Steam Input](https://github.com/HandHeldLegend/SInput-HID). Addressing the major probles that third-party gamepad faced, allowing Analog Joystick and Motion Sensors at the same time while being a bespoke Controller Type.

https://docs.handheldlegend.com/s/sinput/doc/sinput-hid-protocol-dev-SNSaEw36nc

Since then: it's already merged onto mainline SDL and SteamInput, so it all boils down to the end-user implementing the protocol. it might be possible to fully replace or extend "HID USB" with SInput HID instead.

NickGuyver:
adds sinput mode from [HHL](https://docs.handheldlegend.com/s/sinput/doc/sinput-hid-protocol-dev-SNSaEw36nc)
fixes #1438


NickGuyver:
image
image

arntsonl (approved):

PR #1611, opened 2026-02-17
Cleanup GPButton State Logic

j-wut:
Deleting all of the pin mapping logic, and using gamepad's button state instead.

I assume this was a relic of old code, and was never refactored, since I don't see a reason why we would be using the pin state instead of just the gamepad's button states.

This fixes display issues with the HE Trigger Addon, and any other addons that modify gamepad button state.

Tested with HE Trigger Addon, as well as the basic GPIO button mapping.

### DIFF
--- a/src/display/ui/elements/GPButton.cpp
+++ b/src/display/ui/elements/GPButton.cpp
@@ -29,104 +29,21 @@ void GPButton::draw() {
         baseY = ((this->y) * scaleY + this->getViewport().top);
     }
 
-    bool pinState = false;
     bool buttonState = false;
     bool turboState = false;
     uint16_t state = 0;
-    int16_t setPin = -1;
     int32_t maskedPins = 0;
-    bool useMask = false;
-    GamepadButtonMapping *mapMask = NULL;
-    GpioMappingInfo* pinMappings = Storage::getInstance().getProfilePinMappings();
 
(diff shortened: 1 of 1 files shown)


mikepparks:
Please do not try to prematurely optimize code without knowing it's intent. There are a few board-specific layouts that depend on the pin mapping rather than button masks.

j-wut:
> Please do not try to prematurely optimize code without knowing it's intent. There are a few board-specific layouts that depend on the pin mapping rather than button masks.

Are you referring to line 91, or the sections to get pinmask?

mikepparks:
I'm referring to the assumption that we're not using the pin state.

j-wut:
ok, i previously forgot  to include configs/ folder when i grepped for `GP_ELEMENT_PIN_BUTTON`.

For the other inputTypes (`GP_ELEMENT_BTN_BUTTON` and `GP_ELEMENT_DIR_BUTTON`) what was the reason for using mask to get pinstate, instead of directly using gamepad state?

This was my reasoning for wanting to remove the button masking:

| buttonState | pinState | state | |
| ----------- | ----------- | ----------- |----------- |
| False |  False| False | |
| False |  True| False | should use `GP_ELEMENT_PIN_BUTTON` if the button activation was wanted for this case |
| True   | False        | False | This is the case for addons that modify gamepad state (anything other than GPIO mapping) |
| True | True | True |  |

PR #1613, opened 2026-02-25
small keyboard fixes

### LINKED ISSUE #1543: I would like to be able to map the keyboard's screenshot key.
In third-party screenshot software, one need only press PrtSc to capture the screen perfectly. However, I have been unable to locate this key mapping on the GP2040 console. It may be an oversight on my part, but should it indeed be absent, I would appreciate its inclusion in a future firmware update.

Image

Image

NickGuyver:
reordered keycodes by value to make searching easier 
fixed and added some keycodes
small bug in driver
fix #1543


arntsonl (approved):
That might actually be the multimedia keycode fix!!

PR #1617, opened 2026-03-06
I/O Expander add-on modification

kolabo22:
For some reason (I'm currently making a joystick with a large number of additional buttons),
the current specifications of the I/O Expander add-on don't allow me to do anything, so I had no choice but to modify the I/O Expander and sound.
In terms of the changes, I've made it possible to add A3, A4, and EXT1-12 to the I/O Expander button settings in addition to the basic settings, and I'm also thinking of making them usable as macro buttons. The sound is still in the process of being modified. I haven't confirmed all of the functionality, so I apologize if there are any bugs.
By the way, this is a collaboration with Gemini.

3/13 I changed the Wii add-on to I2C0 so that it can be used with an OLED (including other minor adjustments like plugging and unplugging). It can also be used with an I/O expander.

訳有って(大量の増設ボタン付きジョイスティックを製作中)、
今の仕様のI/Oエキスパンダーアドオンでは、どうにも出来ないので、仕方が無いから、I/Oエキスパンダーとサウンドを改造しました。
内容としては、I/Oエキスパンダーのボタン設定に基本設定の他にA3・A4・EXT1～12追加出来る様にし、一応マクロボタンとしても使える様にした心算です。サウンドは改造途中です。全ての動作を確認した訳では無いので、バグが有ったら御免なさい。
因みにジェミニ君との合作です。

3/13 WiiアドオンをI2C0に変更しOLEDと併用可能にしました(抜き差し対応等細々した所も…)。後、I/Oエキスパンダーも併用可能です。

タイトルなし_
スクリーンショット (35)

### DIFF
--- a/src/addons/wiiext.cpp
+++ b/src/addons/wiiext.cpp
@@ -291,6 +291,19 @@ void WiiExtensionInput::update() {
             accelerometerZ = wii->getController()->motionState[WiiMotions::WII_ACCELEROMETER_Z];
             isAccelerometer = true;
         }
+     // --- 共通デッドゾーン処理を追加 ---
+        uint16_t dz = 1000; // LEDが消えない場合はこの値を大きく（例: 2000）してください
+        uint16_t mid = joystickMid;
+
+        if (abs((int)leftX - (int)mid) < dz) leftX = mid;
+        if (abs((int)leftY - (int)mid) < dz) leftY = mid;
+        if (abs((int)rightX - (int)mid) < dz) rightX = mid;
+        if (abs((int)rightY - (int)mid) < dz) rightY = mid;
+
+        // トリガーの微小な誤反応もカット
+        if (triggerLeft < 10) triggerLeft = 0;
+        if (triggerRight < 10) triggerRight = 0;
+        // ---------------------------------
     } else {
         currentConfig = NULL;
     }


mikepparks:
I'm going to advise you to revisit this PR and break it up into multiple PRs. There are a lot of changes here that are out of scope for the PR as titled, and would be difficult to approve on its own.  Also, our code standard typically expects all comments to be made in English. I would like to see these changes come back in separate PRs to make it easier to know what's changed for what purpose, and not combine everything into one. Thanks.

mikepparks:
Also, when you update the Wii code, please do not remove the ability to choose I2C blocks. That is a design decision and needs to remain there. It is important that users are able to choose their I2C blocks, and the code will auto-detect which devices are on which block when they're enabled.

kolabo22:
So, I should add English to the comments and fork it as a separate topic, right?

In this case, it would be best to separate it into I/O expansion and Wii modification, right?

However, with the current Wii add-on, simultaneous use with the OLED display was impossible, so I changed the channel. Would it be better to make it selectable on the web configuration?

I was working on creating a joystick with as many features as possible, which is why I'm tinkering with it.

コメントに英語をプラスして、別項目でフォークすればいいのですね。
この場合はI/O拡張とWii変更と言う感じで分けたら良いのですね。

だだ現在のwiiアドオンだと、OLEDとの併用が不可能だったので、チャンネル変更を行った次第です。WEBコンフィグ上で選択出来る様にしたら良いでしょうか?

出来るだけ機能を搭載した、ジョイスティックを作成中だったもので、その為弄っている次第です。
DSC_0632

mikepparks:
Yes, keep the two separate. If the Wii code fixes the OLED, that should be one PR, and the IO Expander a second PR.  Ideally, you should be able to have the OLED on I2C0 and the Wii on I2C1 at the same time, and both will work as long as both are enabled.  They're two separate blocks, and the code will automatically scan each block to see which devices are connected.  I would usually suggest having the Wii and IO expander on the same block, rather than the Wii and OLED, since the OLED likes being the only device on the block.

ご理解いただきありがとうございます。

kolabo22:
After various tests, the wiring is indeed as you pointed out.
The OLED channel is wired separately using I2C0 (connecting the OLED and I/O expander simultaneously caused the I/O expander to malfunction).
The I/O expander and Wii controller were wired using I2C1, and it worked without any problems. The reason I fixed the Wii add-on's I2C address was because the controller wasn't recognized correctly when set to automatic. The modified joystick shown in the photo works fine. However, I don't own all Wii peripherals, so I can only test with the Nunchuk and controller...

I will organize everything and submit another pull request.

I apologize for the trouble. Thank you very much for your help.


色々検証した結果、ご指摘の通りの配線に成っています。
OLEDのチャンネルは、I2C0で単独にて配線し、(OLEDとI/Oエキスパンダーを同時に接続するとI/Oエキスパンダー側が誤動作を起こしました。)
I/OエキスパンダーとWiiコントローラー配線はI2C1で配線した所、異常は発生せずに旨く動作しました。WiiアドオンのI2Cアドレスを固定したのも、自動だと上手くコントローラーが認識されなかった為です。写真で紹介した改造ジョイスティックでは、旨く動作しています。ただ、Wiiの全ての周辺機器は流石に持って無いので、ヌンチャクとコントローラーまでしか検証は出来ませんが…

色々整理してから再度プルリクエスト致したいと思います。
お手数をお掛け申し訳有りません。この度は有難うございました。

DSC_0648

Issue #1425, opened 2025-07-14
Use GPIO pins for boot input mode select


rmaccrimm:
### Problem

After changing profiles, the physical buttons used to select input mode may change on the next boot due to different pin mappings. Often I just end up doing a lot of guess-and-checking as I restart the device until I get the right input mode.

### Solution

I've implemented a fairly simple alternative in which you enable a GPIO input mode select flag and then configure a list of mappings from pin numbers to input modes. It adds the following new fields to GamepadOptions

<img width="641" height="151" alt="Image" src="https://github.com/user-attachments/assets/4e28d054-aa3c-4244-b8ea-5e15831946d8" />

### Do you plan to work on this feature yourself?

Yes, I've already got the basic functionality working. If there aren't any problems/concerns around adding this I can figure something out for the web config.

TheTrainGoes:
The is an interesting thing, also something that we have talked about before not just from the booting perspective but also other shortcuts and the FN layer.  

Overall I agree that this is a net-positive change and something we can look at less as connecting boot modes to buttons and more as looking at them as connected to GPIO which is more consistent with our current setup.

PR #1620, opened 2026-03-17
GPIO-based Input mode and profile select on boot

rmaccrimm:
This PR implements a couple of related features I've been working on:

1.  #1425
2. Option to select a profile to load along with the input mode

My goal with the second was to be able to set per-console profiles but not require a hotkey to be set to use them (to reduce the risk of accidentally triggering the hotkey mid-game).

This is accomplished with new configuration separate from the existing boot input-mode mapping, called `BootModeOptions`. It includes a flag to enable/disable the functionality and I've done my best to maintain the old behaviour when it's disabled. The web-config looks like this (it's an entirely new page as that was just easier, largely based off the pin mapping page):

Screenshot 2026-03-17 at 11 38 03 AM

Some other non-functional changes I made:
- Reworked the BootAction type to try to reduce the number of switch statements to update
- move the INPUT_MODES object from SettingsPage.jsx to a new file so it would be easier to re-use on a new page
- changed what appears to be a random typo I came across (REVERSE_UP_DEFUALT)


arntsonl (dismissed):
This is a really cool change, thank you for adding this! Code looks good, I'll have to verify all of our boot modes for individual boards are setup correctly on boot, but we can add that as a new PR if we need new fixes.

PR #1625, opened 2026-04-01
Add GPIO28 and 29 to I2C block 0

Vixima:
Allows the user to select GPIO pins 28 and 29 for I2C block 0 in the web configuration tool, Waveshare Pico Zero specific change since said pins are to easier to access (depending on mounting) compared to pins 17 to 25.


arntsonl (approved):
Didn't even know it could do this, but I'm getting the thumbs up

PR #1628, opened 2026-04-20
Assignable Hotkey Actions for web-config and USB mode

mortencombat:
I'm building a joystick with only 3 physical buttons (B1, B2 and Fn) and would like to be able to access both web-config and USB mode without having to open the case and access the physical buttons on the RP2040 Mini Breakout Board I am using.

There is already a hotkey action for **Reboot GP2040** (reboots into default/controller mode). This PR adds hotkey actions for **Reboot GP2040 into web-config mode** and **Reboot GP2040 into BOOTSEL/USB mode**.

Build tested and working as intended on the RP2040 Mini Breakout Board v7.5E Passthrough Edition.

### DIFF
--- a/proto/enums.proto
+++ b/proto/enums.proto
@@ -461,6 +461,8 @@ enum GamepadHotkey
     HOTKEY_RS_DOWN               = 83;
     HOTKEY_RS_LEFT               = 84;
     HOTKEY_RS_RIGHT              = 85;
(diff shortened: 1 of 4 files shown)


arntsonl (approved):
Totally good with this! Hotkeys for web-config/usb boot

PR #1631, opened 2026-04-25
Add ADS1115 External ADC support

Marckus5:
Introduce support for the ADS1115 external ADC, including functionality for per-channel enable, inverted controls, and deadzones (no way to change them in the web config yet).


arntsonl:
Merging this one in, we can tweak the deadzone stuff a bit for the ADS1115 but this gives us better flexibility with the ADS1219 vs. ADS1115 chips.

PR #1632, opened 2026-04-27
implement mDNS support and hostname configuration

Pelsin:
Screenshot 2026-04-27 at 22 16 50

Tested on macOS, windows 11, and iOS
Also possible to change hostname in boardconfig


arntsonl (approved):
Looks clean! I like that we hard-code the DNS but if someone wanted to make their own build with a special domain they could

PR #1633, opened 2026-04-27
Fixed REVERSE_X_DEFAULT typo in identifier

speedypotato:
2 liner typo fix.  I ran into this compilation issue when trying to compile a custom config and the compiler identified two places where the identifier was being defined.


arntsonl (approved):
Yup! This is a bug

PR #1635, opened 2026-04-29
use the original dpad input, not the modified dpad output, in hotkeys

### LINKED ISSUE #1634: dpad in LS/RS mode no longer trigger dpad-based hotkeys
#1593 introduced an innocuous change which makes sense but is incomplete holistically. To let addons (such as the SNES input addon) trigger hotkeys, `gamepad->hotkey()` was moved to after `addons.ProcessAddons()`. Makes total sense --- however, addon processing itself is *also* after `gamepad->process()`, where this (also sensible) code lives:

```
	switch (activeDpadMode)
	{
		case DpadMode::DPAD_MODE_LEFT_ANALOG:
			[...]
			state.dpad = dpadOnlyMask;
			break;

		case DpadMode::DPAD_MODE_RIGHT_ANALOG:
			[...]
			state.dpad = dpadOnlyMask;
			break;

		default:
			break;
	}
```

Essentially, when the dpad (the input control element) is in LS or RS mode, its bits are masked off to only be the digital inputs. This is *good*, because dpadOnlyMask might have come from direct directional buttons, letting you press both dpad-in-LS-mode and direct-directional inputs and get both LS and DP out. Here, essentially, `state.dpad` is being used as the *digital output*. However, code to process hotkeys thinks `state.dpad` is still an *input*:

```
	inline bool __attribute__((always_inline)) pressedDpad(const uint8_t mask) {
		return (state.dpad & mask) == mask;
```

...and with the dpad-as-input bits masked off by that point, this comparison always fails when the dpad is in LS/RS mode. Notably, the comparison now always works with direct directionals, which was not the case before this merge --- it only ever looked at the dpad-as-input state.

Another way, `state.dpad` is overloaded, and represents both an input (what inputs are active on the input control element, the dpad) and an output (what digital directions to send to the host). I propose breaking them up.

All this to say, I suspect `pressedDpad` should look at `state.dpadOriginal`. I will test this when I get some free moments today.

bsstephan:
hotkey processing used to be before the core gamepad process, so it was reading the unmodified dpad input, but #1593 moved hotkey processing to after, so what it ends up reading is the output, which gets bits masked off of it when in LS/RS mode (so as to not do a double output of a direction in LS and DP outputs, when that's likely not intended). all the code involved here makes sense, but it broke in tandem as an order of operations sequence.

this separates the concerns by having the hotkey dpad read use dpadOriginal, which is only used for the display, but is exactly what we need --- the dpad input element, untouched by gamepad output processing. I moved where dpadOriginal gets set in order to make its purpose clearer --- it happens on read(), now, not process()ing.

fixes #1634

### DIFF
--- a/headers/gamepad.h
+++ b/headers/gamepad.h
@@ -69,7 +69,7 @@ public:
 	 * @brief Check for a dpad press. Used by `pressed[Dpad]` helper methods.
 	 */
 	inline bool __attribute__((always_inline)) pressedDpad(const uint8_t mask) {
-		return (state.dpad & mask) == mask;
+		return (state.dpadOriginal & mask) == mask;
 	}
 
 	/**
--- a/src/gamepad.cpp
+++ b/src/gamepad.cpp
@@ -277,9 +277,6 @@ void Gamepad::process()
 		state.dpad = filterToFourWayMode(state.dpad);
 	}
 
-	// hold current dpad state regardless of input
-	state.dpadOriginal = state.dpad;
-
 	// stash digital-only dpad state for later
 	uint8_t dpadOnlyMask = ((state.dpadOriginal & 0xF0) >> 4);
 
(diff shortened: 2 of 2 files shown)


arntsonl (approved):
Another good bug fixed, thank you @bsstephan !

PR #1637, opened 2026-04-30
Update layout manage to fallback if no ALT layout exists

TheTrainGoes:
This PR allows layout manage to fallback if you overwrite a firmware with another version that does not have the available and currently set ALT display layout


arntsonl (approved):
Yep I've run into this before! Good catch

PR #1641, opened 2026-05-05
fix duplicate analog stick mode

NickGuyver:
fix to allow both sticks to be assigned to one mode (right, left)


arntsonl (approved):
Looks good! Easy way to have two sticks for one mode

PR #1646, opened 2026-05-08
Claude/add curvecade layout soudt

dgjohnson:


mikepparks:
This looks like a personal fork which modifies the Pico profile, and will be rejected. Also no PRs with Claude taint will be accepted at this time.

PR #1650, opened 2026-05-09
:bug: Fix Macro D-Pad inputs being dropped & clearly separate physical vs processed D-Pad checks

Niconiconi-OwO:
### Describe the bug
This PR fixes a bug introduced by the recent hotkey D-pad fix (PR #1635 / Issue #1634),addons (such as Input Macros) that inject D-Pad inputs were having their directional outputs completely overwritten and ignored. This occurred because recent hotkey/UI fixes relied on state.dpadOriginal for masking and checking inputs. Since state.dpadOriginal is captured before PreprocessAddons(), any D-pad inputs injected by macros were being erased during Gamepad::process(), and the USB output was blind to them.

### Root Cause
In PR #1635, state.dpadOriginal = state.dpad; was moved to Gamepad::read() to ensure hotkey processing uses unmodified physical inputs. This successfully fixed the hotkey issues.
﻿
However, Gamepad::process() was still using state.dpadOriginal to calculate dpadOnlyMask and dpadModeMask.
The execution order in the main loop is:

1. read() -> state.dpadOriginal is set (contains ONLY raw physical inputs).
2. ﻿PreprocessAddons() -> Addons (like Macros) inject D-pad inputs into state.dpad.
3. ﻿process() -> Uses state.dpadOriginal to create masks, and then executes state.dpad = dpadModeMask;.

﻿
Because state.dpadOriginal does not contain the Addon-injected inputs, state.dpad = dpadModeMask; completely overwrites and erases any D-pad modifications made by macros/addons.

### The Fix
This PR cleanly separates the "processed game output" from the "raw physical input". It fixes the macro direction issue while keeping hotkeys, LEDs, and UI strictly tied to physical presses:

gamepad.cpp: Introduced a local currentDpadSnapshot in process() to calculate dpadOnlyMask and dpadModeMask. This carries macro-injected D-Pad states safely through the masking pipeline without erasing them.

gamepad.h:

Restored pressedDpad() to return state.dpad (so the USB driver outputs the correct macro inputs).

Added a new pressedDpadPhysical() method returning state.dpadOriginal.

Updated pressedHotkey() to use pressedDpadPhysical().

reactiveleds.cpp & GPButton.cpp: Updated the Reactive LEDs and OLED display to explicitly use pressedDpadPhysical().

### Result

Macros injecting D-Pad inputs now correctly output to the game.

Hotkeys continue to safely ignore macro inputs (preventing accidental mode switches).

Reactive LEDs and OLED displays continue to honestly reflect only raw physical button presses, preserving the original intent of previous display/hotkey fixes.

### Testing
Compiled and tested locally. Verified that macros fire correctly, hotkeys work as intended, and displays/LEDs react only to physical inputs.

### DIFF
--- a/headers/gamepad.h
+++ b/headers/gamepad.h
@@ -69,6 +69,13 @@ public:
 	 * @brief Check for a dpad press. Used by `pressed[Dpad]` helper methods.
 	 */
 	inline bool __attribute__((always_inline)) pressedDpad(const uint8_t mask) {
+		return (state.dpad & mask) == mask;
+	}
+
+	/**
+	 * @brief Check for a raw, physical dpad press. Unaffected by macros, SOCD, or D-pad modes. Typically used for hotkey detection.
+	 */
+	inline bool __attribute__((always_inline)) pressedDpadPhysical(const uint8_t mask) {
 		return (state.dpadOriginal & mask) == mask;
 	}
 
@@ -84,7 +91,7 @@ public:
 	 */
 	inline bool __attribute__((always_inline)) pressedHotkey(const HotkeyEntry &hotkey) {
 		return (hotkey.action != 0 && pressedButton(hotkey.buttonsMask) &&
-				pressedDpad(hotkey.dpadMask) && pressedAux(hotkey.auxMask));
+				pressedDpadPhysical(hotkey.dpadMask) && pressedAux(hotkey.auxMask));
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


xC0000005:
This fixes the root cause of what was breaking my legends otg usb host support, since what I did to fix it was revert the referenced fix.

arntsonl (approved):
Working great, thank you for following our naming conventions!

PR #1655, opened 2026-05-16
Expanding GPIO from 30 to 48 pins to accommodate RP2350B variants

mikepparks:
This PR will expand the storage of GPIO pin mappings from 30 to 48 pins to support the RP2350B controller variant.  Currently, as an example, the Pimoroni Pico Plus 2 config uses this chip, though only a few extra pins beyond the initial 30 are exposed by the board.  The code will adapt to any changes to the `NUM_BANK0_GPIOS` definition in `pico-sdk`, so adding support for this controller will be dependent on the CMake file definitions in your config's directory.


.github/workflows/cmake.yml
```diff
@@ -50,6 +50,7 @@ on:
           - OSUMGP-RP2040
           - PicoAnn
           - PicoFightingBoard
+          - PimoroniPicoPlus2
```
arntsonl, line 53: Do we also want to add WaveshareRP2350BPlusW?

arntsonl (changes requested):
I think we just add the WaveshareRP2350BPlusW to the build and this should be ready!

.github/workflows/cmake.yml
mikepparks: I have that in my next commit.

mikepparks:
This is only holding on wiring up web config for 48 pin support. It's a bit of a lift since many areas are hard-defined for 30 GPIO, and make assumptions about where analog GPIO are located in relation to those differences.

arntsonl (dismissed):

arntsonl (approved):

PR #1659, opened 2026-05-29
Fix: run input driver processAux() before addons on Core 1

JTCruzado:
## Summary

On Core 1, `GP2040Aux::run()` calls `addons.ProcessAddons()` before
`inputDriver->processAux()`. When an external-authentication driver
(P5General) services its USB dongle handshake inside `processAux()`, a
blocking display I²C write inside `ProcessAddons()` (~25 ms for a full
SSD1306 framebuffer at the 400 kHz default) stalls the auth pipeline.
`hash_pending` stays set, so Core 0's `process()` returns early and HID
reports to the PS5 are delayed by 1–2+ frames.

This is the root cause behind the note in #1546: *"I2C0 must be
disabled; otherwise the report rate will be extremely low — cause not
yet identified."*

This PR reorders the Core 1 loop to service `processAux()` **before**
`ProcessAddons()`, so the auth handshake gets a turn at the start of
every iteration and is no longer gated behind the display write.

## Why this is safe

- Pure scheduling change — no new state, no new code paths.
- `ProcessAddons()` still runs every iteration, immediately after `processAux()`.
- Drivers without an auxiliary step inherit a no-op `processAux()`, so
  behavior is unchanged for XInput, PS4, Switch, Switch Pro, Xbox One, etc.
- No shared state between the auth pipeline and the addons it's reordered against.

## Testing

Validated on hardware with a P5General dongle on a PS5:
- Haute42 COSMOX (mine)
- RP2040 Advanced Breakout Board – USB Passthrough (community-validated)

In testing, this scheduling change — combined with a 30 fps display cap
and a 1 MHz I²C bump (submitted as separate follow-up PRs) — brought
OLED-enabled input latency to parity with running the OLED disabled:
single-frame taps register as single frames and quarter-circle motions
land consistently. I'm submitting the scheduling fix first as the
foundational, most universal piece; the auth pipeline can only benefit
from running earlier, and the follow-ups stack on top.

## Follow-ups (separate PRs)
- Configurable display frame interval (default 30 fps) to reduce Core 1
  occupancy by the display addon.
- Per-board I²C0 Fast Mode Plus (1 MHz) for boards with an SSD1306/SH1106
  OLED + USB auth.

### DIFF
--- a/src/gp2040aux.cpp
+++ b/src/gp2040aux.cpp
@@ -52,13 +52,21 @@ void GP2040Aux::setup() {
 
 void GP2040Aux::run() {
 	while (1) {
-		// Pre, Process, and Post
 		addons.PreprocessAddons();
-		addons.ProcessAddons();
 
-		// Run auxiliary functions for input driver on Core1
+		// Run auxiliary input driver functions (e.g. P5General auth USB
+		// listener servicing hash_pending round-trips to the dongle)
+		// BEFORE ProcessAddons() so the auth pipeline gets a turn at the
+		// start of every Core 1 iteration. Otherwise a long display I2C
+		// write inside ProcessAddons() (~10 ms at 1 MHz, ~25 ms at 400 kHz)
+		// stalls the auth listener, causing hash_pending to stay set and
+		// delaying HID reports to the PS5 by 1-2+ frames. Drivers without
+		// an auxiliary step provide a no-op processAux(), so this ordering
+		// is safe for all driver modes.
 		if ( inputDriver != nullptr ) {
 			inputDriver->processAux();
 		}
+
+		addons.ProcessAddons();
 	}
 }


arntsonl:
I've looked at this a few different ways and I'm still pretty confused on the timing order.

Essentially, we reduce the I2C0 timing with using the 1 MHz fast mode plus which has a viable impact.

Display frame interval hindering is basically de-synchi'ng the display to the new fast mode. So it skips a few frames which is good! But we are not displaying that quickly so it doesn't make sense to restrict the display mode that way.

Re-ordering process() and usb host code, there is no sleep or wait type of code. We are firing as fast as possible. So basically the code right now is:

preprocess
process (big display delay)
usb host code
preprocess
process (big display delay)

this changes it to

preprocess
usb host code
process (big display delay)
preprocess
usb host code
process (big display delay)

so the delay is still there, and I don't believe the pre-process has any impact on display.

so... I think the only one of these 3 check-ins with an impact is the fast-mode on I2C0 after looking at this for a few weeks.

arntsonl:
Going to close this one and the display fps, then leave the fast mode open.

PR #1669, opened 2026-06-18
scrolling text in status bar

f3bandit:
scrolling status bar layout patch
increase macro limit to 15 patch
layout fix for button layout after scrolling status bar patch


TheTrainGoes:
Hi @f3bandit ,

PRs should be broken out and submitted individually based on what you are adding.  

We do not accept patch files.

We will not up the macro limit to 15 as it will take a large amount of space to do so.  

The first and last one look to be part of the same thing so they should be able to be submitted together.

PR #1672, opened 2026-06-24
Gamepad USB Host Rework

arntsonl:
This is a rework of the Gamepad USB Host

This breaks up all inputs into their own separate .h/.cpp in the parent /hosts/ folder

Additionally, this breaks up VID/PID parsing into if() switch clauses to capture multiple devices of the same product maker


arntsonl (commented):
Don't see any debug code left, should be parity with our gamepad host!

PR #1678, opened 2026-07-11
Refactor runSOCDCleaner for cleaner logic

yemihai:


mikepparks:
Removing features is not cleaner logic. You removed essential checking for SOCD modes.

PR #1682, opened 2026-07-19
Extend the range of characters for profile names

### LINKED ISSUE #1663: Allowing Profile Labels to use more characters in the extended ASCII range.
### Problem
Profile Labels are currently limited to only Letters, Numbers, and Spaces on the Web Gui. It seems like this limitation is overly strict and could be changed to allow more characters. This currently prevents users from adding basic special characters to a label. 

Changing this would let you create unique and silly titles like: DR+DI SLOP!!, Tekken (-_-), GG+R.AcCore#XRD

### Solution
Increase the range of allowed characters to include ranges like Dec 33-47, 58-64, 91-96, and 123-126 (Potentially 128-254 as well) I used this for reference: [https://www.asciitable.com](https://www.asciitable.com)

Image

I did have a discussion about this in the discord server, just to make sure there is no actual technical limitation that prevents this from being possible, and it sounded like there wasn't.
https://discord.com/channels/1049366310389289001/1196548991513538680/1513692307193593928

If no one else wants to change it, I might? It is probably easiest for someone else to do it who has contributed before since I would have a lot of extra work to do to add such a seemingly small change.

TheTrainGoes:
This PR extends the usable characters for the profile name field from `a-zA-Z0-9` to `\x20-\x7e` which includes, I believe, all of the normally used characters.

Please note that this was a quick one line fix after reading about it but I did use Claude to do a quick translation on the web-config stuff to other languages.

### DIFF
--- a/www/src/Locales/en/PinMapping.jsx
+++ b/www/src/Locales/en/PinMapping.jsx
@@ -6,7 +6,7 @@ export default {
 	'pin-pressed': 'Pressed GPIO Pin: {{pressedPin}}',
 	'profile-label-title': 'Profile name',
 	'profile-label-description':
-		'Max 16 characters. Letters, numbers, and spaces allowed.',
+		'Max 16 characters. Printable ASCII characters allowed.',
 	'profile-pin-mapping-title': '{{profileLabel}} - GPIO Pin Mapping',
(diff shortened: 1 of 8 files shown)


arntsonl (approved):
Looks like an easy fix!

PR #1686, opened 2026-07-27
Fix PS4 sensor data incorrect descriptor and alignment

DvdGiessen:
The PS4 sensor data descriptor started with an incorrectly sized field, causing the data to be offset by a single byte. The byte order swap in the driver was presumably added after a observation that the fields seemed in a different byte order, while actually it was the bytes from two different fields being combined.

This makes the gyro/accelerator work correctly, as well as all the other fields that follow them.

Note: I haven't actually tested this with one of the already supported IMU drivers (I think it's just the Wii?), only with my custom build using a QMI8658 (see #1161).

(Fun fact, not spotting this earlier was the reason I never finished that PR. With this fixed it actually works quite well now, and I hope it may also help others working on other IMUs.)

### DIFF
--- a/headers/drivers/ps4/PS4Descriptors.h
+++ b/headers/drivers/ps4/PS4Descriptors.h
@@ -135,7 +135,7 @@ typedef struct __attribute__((packed, aligned(1))) {
 } PSSensor;
 
 typedef struct __attribute__((packed, aligned(1))){
-    uint16_t battery;
+    uint8_t temperature;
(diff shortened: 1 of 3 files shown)


arntsonl:
Oh nice catch on this one!! If this fixes your old PR, feel free to re-open it. You'll probably need to pull in a lot of changes, so might be worth just making a new branch.

arntsonl (approved):
Very cool!

PR #1688, opened 2026-07-31
Add pvs-studio blurb

NickGuyver:
https://pvs-studio.com/en/order/open-source-license/


arntsonl (approved):
Yeah probably good to include these since we're doing some bigger clean-ups :)

PR #1689, opened 2026-08-10
Fix Haute42|COSMOX C-AT config

TheTrainGoes:
This PR fixes the Haute42|COSMOX C-AT config.  There was an issue with the case LEDs not being set to a long enough number.


arntsonl (approved):
Easy fix!

PR #1690, opened 2026-08-10
debounce updates

NickGuyver:
fixed a couple small bugs
changed debounce loop to only check changed gpio instead of scanning all


arntsonl (approved):
Nice, easy off by one fix. So our default for debounce was 6ms instead of 5ms, which is still well within our margins.

PR #1691, opened 2026-08-11
Add Host Lighting add-on: live RGB LED control from host software over a vendor HID interface

djGLiTCH:
## Summary

Host Lighting is an optional add-on exposing a second vendor HID interface
(usage page 0xFF47, fixed 64-byte reports) that lets host software drive
the board's RGB LEDs while the controller interface operates normally.
Hosts can address individual buttons, the case strip and player LEDs, in
whole-frame or overlay mode; the board's own animations resume when the
host releases control, disconnects, or goes quiet past a keepalive
timeout.

The board serves its own LED map, colour order, identity, runtime state
and animations over the same interface, read from the live configuration -
no per-board data exists on the host side, and user remaps are picked up
automatically. Discovery is by the vendor usage page, which is stable
across input modes and VID/PID overrides.

Disabled by default like all add-ons; while disabled, every mode presents
byte-identical descriptors to stock firmware.

## Trying it

CI artifacts from this PR contain a UF2 per board config. Flash one,
enable the add-on under Configuration -> Add-Ons -> Host Lighting, and the
lighting interface enumerates alongside the controller.

Prebuilt UF2s for every board, from this PR's exact commit, are also
published here if that is easier: https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.0
The same release carries a set built on top of the LED refactor (#1514)
for anyone testing that combination, since CI does not build those.

Python reference clients are attached below (discovery, capability decode,
streaming, per-button tests); hlp-ping.py runs a conformance pass
against a connected board. The attachment is self-contained for quick
testing against this PR; OpenStickCommunity/gp2040ce-binary-tools#12
proposes packaged, cross-platform versions of the everyday tools as
their permanent home.

HLP Python Reference Clients v1.0: [hlp-tools_HLP-v1.0.zip](https://github.com/user-attachments/files/30922356/hlp-tools-attachment.zip)
HLP Python Reference Clients v1.1: [hlp-tools_HLP-v1.1.zip](https://github.com/user-attachments/files/31156958/hlp-tools_HLP-v1.1.zip)
HLP Python Reference Clients v1.2: [hlp-tools_HLP-v1.2.zip](https://github.com/user-attachments/files/31156951/hlp-tools_HLP-v1.2.zip)
HLP Python Reference Clients v1.3: [hlp-tools_HLP-v1.3.zip](https://github.com/user-attachments/files/31156920/hlp-tools_HLP-v1.3.zip)

## Design notes

- Protocol 1.0 with a written compatibility contract: command IDs sit in
  function ranges with reserved slots; from 1.0, IDs and payload layouts
  are frozen and minor versions are additive-only. Full reference in
  docs/host-lighting.md (in this PR).
- Stock descriptors are untouched; descriptor variants are selected at
  runtime, and CFG_TUD_HID is already 2. Console modes (PS3/PS4/PS5,
  Switch, Xbox) never expose the interface.
- XInput behaviour is configurable (AUTO/ON/OFF, default AUTO). AUTO
  boots with the stock console-identical identity and re-enumerates as
  the composite only if the host sends no console authentication within
  ~4 s; the composite binds the OS Xbox 360 driver (MS OS 1.0 compatible
  ID on Windows, xpad vendor match on Linux). Verified against a real
  Xbox 360 console (see Validation). ON forces the composite; OFF keeps
  XInput stock.
- Hosts edit a staged frame; the COMMIT command publishes it atomically
  to the render core through a sequence-locked double buffer, so
  multi-report frames cannot tear.
- Config writes go through the existing deferred GPStorageSaveEvent path;
  nothing writes flash from USB context. SET_INPUT_MODE validates against
  an explicit mode list; unknown modes are rejected rather than rebooted
  into.

## LED refactor (#1514)

Ported experimentally onto #1514's head: 13 of the 15 modified files and
all new files applied verbatim; the adaptations needed were the
render-hook line and the function reading the light layout for
capabilities. The full test suite passed on that pipeline at the same
throughput. On the refactor, the positions capability page carries real
per-light coordinates (empty on the classic pipeline), and controls with
multiple wired LEDs light all of them. I can rebase this PR onto the
refactor once it is accepted into main, or hold this PR until then.

## Validation

Hardware: Haute42 COSMOX (16 LEDs) and COSMOX M-Ultra (46 LEDs including
a 30-LED case strip), both render pipelines, plus a real Xbox 360 console.

- Conformance: protocol 1.0 reported, all command IDs recognized, unknown
  commands rejected, capabilities matching each board's wiring (case
  strip included; per-light positions on the refactor pipeline).
- Streaming: 300/300 commits acknowledged at a sustained 60 fps (three
  staging reports plus COMMIT per frame); overlay compositing over live
  animations; timeout fallback; input-mode round-trips.
- Disabled state: stock 045E:028E identity on the OS driver, no lighting
  interface on the bus, auto-detect never fires.
- Xbox 360 console: AUTO keeps the stock identity, no re-enumeration,
  inputs behave as stock.
- Multi-board: two boards enumerate as independent devices with distinct
  factory IDs and were driven simultaneously without cross-talk.
- Input latency versus stock main (5fb6404d), same board and session,
  identical loop-counter probe: ~0.2 us added to the input sampling
  interval at 60 fps and ~0.3 us at 100 fps (~0.03 % of the 1 ms USB
  polling budget); no configuration measured slower than stock; the cost
  was unchanged on the board needing twice the USB traffic per frame;
  enabled-but-idle is not measurable. Rendering stays capped by the
  existing 100 Hz LED tick.
- All board configs compile for RP2040 and RP2350, both pipelines.
- Development approach and tooling were discussed with the team on
  Discord ahead of this PR.

## Scope

- The XInput commit and the web UI commit are self-contained and can be
  split out if that makes review easier.
- English locale only; other locales fall back.
- Companion documentation PR: OpenStickCommunity/Site#119

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ src/display/GPGFX_UI.cpp
 src/drivermanager.cpp
 src/eventmanager.cpp
 src/layoutmanager.cpp
+src/hostlighting.cpp
 src/peripheralmanager.cpp
 src/storagemanager.cpp
 src/system.cpp
--- a/docs/host-lighting.md
+++ b/docs/host-lighting.md
@@ -1,0 +1,274 @@
+# Host Lighting
+
+Host Lighting lets software on a connected PC drive a GP2040-CE board's RGB
+LEDs in real time over USB, alongside the normal controller function. Game
+state, per-button effects, ambient scenes - anything a host application can
+compute, it can show on the board's lights, while the on-board animations
+take over automatically the moment the host goes quiet.
+
+The feature is fully self-describing: hosts query each board's LED layout,
+identity and state over the same interface, so host software needs no
+per-board configuration or updates when new boards are released.
+
+## How it works
+
+The add-on is disabled by default, like all GP2040-CE add-ons. Enable it in
+the web configurator (`Configuration -> Add-Ons -> Host Lighting`); board makers
+can ship it on by defining `HOST_LIGHTING_ENABLED 1` in a board config.
+
+When enabled, supported input modes expose one extra vendor-defined HID
+interface next to the regular controller interface. Hosts exchange fixed
+64-byte reports on it:
+
+- Lighting commands write into a **staged frame** on the board; a `COMMIT`
+  publishes it atomically to the LED render loop, so multi-report frames never
+  tear.
+- The render loop shows the host frame while it stays fresh. If the host
+  releases control - or simply stops talking for a configurable timeout
+  (default 2 s) - the board's own animations resume seamlessly. A crashed or
+  disconnected host can never leave the lights stuck.
+- **Whole-frame** takeover replaces all lighting; **overlay** takeover only
+  replaces pixels the host has explicitly staged, compositing host effects
+  over the running animations.
+- Everything the host needs to know - LED map, colour order, board identity,
+  current state - is served by `GET_CAPS` from the board's live configuration,
+  including user remaps.
+
+## Supported modes
+
+| Input mode | Lighting interface |
+|---|---|
+| Generic (DInput) | Yes |
+| Keyboard | Yes |
+| XInput | Yes - see below |
+| PS3/PS4/PS5, Switch, Xbox, other console modes | Never |
+
+**XInput** carries a choice, configurable in the web configurator
+(`Host Lighting` section):
+
+- **Auto** (default): the board boots with its stock, console-identical
+  identity. A console begins authentication immediately and the board stays
+  stock - consoles never see any difference. A PC sends no console
+  authentication, so after ~4 seconds the board re-enumerates as a composite
+  whose controller interface binds the operating system's own Xbox 360 driver
+  (Windows via MS OS descriptors, Linux via the kernel's xpad vendor match),
+  with the lighting interface alongside. Hosts should expect the lighting
+  device to appear a few seconds after plug-in in this mode.
+- **Always on**: the composite identity from boot (PC-only while enabled).
+- **Off**: stock XInput identity always; no lighting interface in XInput mode.
+
+With the add-on disabled (the default), every mode presents byte-identical
+USB descriptors to stock firmware.
+
+## Discovery
+
+Do not match VID:PID - it varies by input mode and user override. Enumerate
+HID devices and match the top-level collection:
+
+- **Usage page `0xFF47`, usage `0x4C`**
+- One 64-byte input report and one 64-byte output report, no report IDs
+  (Windows buffers are 65 bytes with a leading `0x00` report-ID byte)
+
+Then send `PING` and require the `GPHL` magic and protocol version >= 1.0.
+Bind boards persistently by the factory-unique ID from `GET_CAPS` page 0 -
+never by device path, which changes with USB ports.
+
+Multiple boards connected at once appear as fully independent lighting
+devices - one interface per board, each addressed separately with no
+interaction between them.
+
+## Protocol reference
+
+The wire protocol carried on this interface is the Host Lighting Protocol
+(HLP) - the `GPHL` magic in `PING` replies and the `hlp-` prefix on the
+reference tools refer to it.
+
+All transfers are 64-byte reports. Requests: `[0]=command, [1]=sequence,
+[2..]=payload`. Replies echo the command with bit 7 set: `[0]=cmd|0x80,
+[1]=sequence, [2]=status, [3..]=payload`. Status: `0` OK, `1` unsupported
+command, `2` invalid argument.
+
+### Commands
+
+| Cmd | Name | Payload | Notes |
+|---|---|---|---|
+*Session and discovery (0x01-0x0F)*
+| 0x01 | PING | - | Reply `[3..6]="GPHL", [7..8]=version`; any OK command refreshes the takeover keepalive |
+| 0x02 | GET_CAPS | `[2]=page` (+`[3]=start` for page 4) | See pages below |
+| 0x03 | SET_MODE | `[2]=takeover (0 whole-frame, 1 overlay), [3..4]=timeout ms LE (0->2000, min 100), [5]=apply board brightness` | Send at connect; settings persist until reboot |
+
+*Frame staging (0x10-0x2F)*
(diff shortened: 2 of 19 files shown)


djGLiTCH:
HLP v1.1, v1.2, and v1.3 are now on the branch, as one commit each. All are additive: no command ID or existing payload moved, the capability format byte stays at 2, and a v1.0 host keeps working against a v1.3 board unchanged.

### HLP Version History

| | v1.0 | v1.1 | v1.2 | v1.3 |
|---|---|---|---|---|
| Released | 11 Aug 2026 | 16 Aug 2026 | 16 Aug 2026 | 17 Aug 2026 |
| Commands | 14 | 14 | 15 (+ `SET_LIGHT` 0x15) | 16 (+ `SET_LIGHT_RGBW` 0x16) |
| Capability pages | 0-4 | 0-5 (+ light table) | 0-5 | 0-5 |
| Page 1 runtime state | mode, profile, brightness, player, fingerprint, animation index | + feature bitmask, LED framework, animation namespace, render Hz | unchanged | unchanged |
| Per-light visibility | one range per control (page 2) | every light named, with owner, GPIO, position and flags (page 5) | unchanged | unchanged |
| Per-light addressing | raw index (`SET_RANGE`) | raw index (`SET_RANGE`) | by page 5 ordinal (`SET_LIGHT`) | + RGBW by ordinal (`SET_LIGHT_RGBW`) |
| Staging reply detail | applied/skipped counts | unchanged | unchanged | + per-entry outcome mask naming each skip |
| Stageable button IDs | 0-17 + specials 24-29 | same (18-19, 30-41 report-only) | every named ID, incl. A3/A4 and E1-E12 | unchanged |
| Fingerprint scope | per-control lookups | whole pin map; also echoed on page 5 | unchanged | unchanged |
| Caps format byte | 2 | 2 | 2 | 2 |
| Detection | `PING` magic `GPHL` | minor >= 1 | minor >= 2 | minor >= 3 |
| Round-trip / 60 fps stream | 2.0 ms / 300 of 300 | 2.0 ms / 300 of 300 | 2.0 ms / 300 of 300 | 2.0 ms / 300 of 300 |

Test builds for every protocol version are published on my fork, each with Classic and Refactor images for all supported boards plus the matching host tools: [HLP v1.0](https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.0), [HLP v1.1](https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.1), [HLP v1.2](https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.2), and [HLP v1.3](https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.3) (v1.3 is the current one to test against).

Gate on `>= v1.0` and read the capability pages; detect the newer surfaces by `PING` minor version, never an exact match.

**v1.1**:
Capability page 5, the light table: one 12-byte record per light (owning button ID, GPIO, action, player index, case group, grid position, flags), paged four per report, fingerprinted. It is the only page that can express a control owning several lights. Page 1 appends a feature bitmask, LED framework, animation namespace and render rate at `[12..18]`, each reserving 0 for "not reported". Also: staged writes are bounds-checked, the fingerprint covers the whole pin map, and the takeover releases on USB disconnect/suspend instead of waiting out the keepalive.

**v1.2**:
`SET_LIGHT` (0x15): stage single lights by page 5 ordinal, up to 15 per report, reply counts applied/skipped; out-of-range ordinals skip, same rule as a control without a light. `SET_BUTTONS` entries naming A3/A4/E1-E12 now apply where such lights exist (previously report-only).

**v1.3**:
Staging replies define `[5..6]` as a per-entry outcome mask (bit n set = entry n applied; popcount equals the applied count), so a skip names the stale ordinal instead of forcing a page 5 re-walk. `SET_LIGHT_RGBW` (0x16): the RGBW pairing `SET_RANGE` already had, 12 entries per report, W ignored on chains without a white channel. On chains with one, host-supplied W now renders: the framework's achromatic shortcut never read W, so subtractive white (0,0,0,W) drew black - hosts send W = min(R,G,B) with RGB reduced, gated on minor >= 3.

**Validation:**
80 assertions per pipeline at v1.3, all passing - every command, every capability page, paged walks, out-of-range and mixed batches, mask correctness including a mid-batch skip in the mask's high byte, and older hosts reading newer boards. v1.1 was rig-tested on a Haute42 B16 and a COSMOX M Ultra Gen 2; v1.2/v1.3 on the M-Ultra, with extended-ID staging proven both ways via temporarily configured A3/E3 lights. Latency re-measured at each version: 2.0 ms median round-trip, 300/300 commits at a sustained 60 fps - nothing added runs per frame.

`docs/host-lighting.md` carries the full reference and changelog. Host tools: OpenStickCommunity/gp2040ce-binary-tools#12; the attached [`hlp-tools`](https://github.com/user-attachments/files/31156920/hlp-tools_HLP-v1.3.zip) zip is updated for v1.3.

djGLiTCH:
I will review this PR again next week and update it to support the latest revised LED refactor code (#1514) now that it has been merged.

djGLiTCH:
Rebuilt on current `main` (`6fe371d3`) now that the LED refactor (#1514) is merged, and force-pushed as a clean series of seven commits (`772982a7..feba721f`). Host Lighting now targets the merged LED pipeline only. The wire protocol is unchanged by the port, and SInput joins XInput, Generic and Keyboard as a mode that can carry the lighting interface.

This also brings the protocol to v1.4, which is additive: capability page 6, the control table, lists every GPIO pin carrying an action, lit or not, so a host can configure its inputs from the board. Details and the changelog are in `docs/host-lighting.md`.

Tested on two boards (46 and 16 LEDs) in all four modes; with the add-on disabled, every mode still presents its stock USB descriptors.

Pre-built firmware for every board config is on my fork for anyone who wants to try it: https://github.com/djGLiTCH/GP2040-CE/releases/tag/HLP_v1.4 (these builds also include #1721).

HLP Python reference clients v1.4: [hlp-tools_HLP-v1.4.zip](https://github.com/user-attachments/files/32455915/hlp-tools-attachment.zip)

I will follow this with a v2.0 consolidation of the wire format in this PR shortly, so the final diff is the thing to review once that is complete.

PR #1697, opened 2026-08-15
elf bug fixes

NickGuyver:
Using codex 5.6 sol, I prompted it to review the arm assembly and work backwards to find where the intention of the source code didn't match. Then I went through and validated the bugs via ghidra. After that, fixed and ran ghidra against each fix to verify that assembly was correct.

macroposition on idle from passing index -1 to macrolist 
remove extra boardledenabled
swap PSCLASSIC_MASK_START and PSCLASSIC_MASK_SELECT 
increase range on brake and gas for wheel xinput type 
off by one for keyboard host modifier
out of range clamping for snowflake screensaver

### DIFF
--- a/src/addons/input_macro.cpp
+++ b/src/addons/input_macro.cpp
@@ -66,7 +66,6 @@ void InputMacro::setup() {
     } else {
         boardLedEnabled = false;
     }
-    boardLedEnabled = false;
     prevMacroInputPressed = false;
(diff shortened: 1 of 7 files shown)


arntsonl (approved):
Looks good! Definitely some mix-ups on ORs, ANDs, and some off by ones.

PR #1702, opened 2026-08-19
remove onebitdisplay

NickGuyver:


arntsonl (approved):
Yep we had it in for legacy, but our driver is much better . Thank you @mikepparks

PR #1705, opened 2026-09-02
issue 484

### LINKED ISSUE #484: SNES Input Add-on: randomly direction is not registered until another button is pressed / direction activated without being pressed
### Prerequisites

- [X] I am running the latest version of the firmware found [HERE](https://github.com/OpenStickCommunity/GP2040-CE/releases)
- [X] I have checked the documentation [HERE](https://gp2040-ce.info/#/) and found no answer
- [X] I checked to make sure that this issue has not already been filed [HERE](https://github.com/OpenStickCommunity/GP2040-CE/issues)
- [X] I am reporting the issue to the correct repository

### Context

* Firmware Version: 0.7.4 (happened also on 0.7.3)
* Is this a custom build?: yes
* System: MiSTer SNES Core (xinput)
* Two different Super Famicom pads were tested

### Expected Behavior

Registering directional input when holding the dpad. Not registering directional input when dpad is released.

### Current Behavior

Sometimes the direction is not registered while holding the dpad. (E.g. right)
When an additional button is pressed then the direction will be registered properly.

Also sometimes mostly right and sometimes left is activated even though it is not pressed/was pressed very shortly and then released.

### Steps to Reproduce

1. Use this configuration. (Setting regular pins to -1 is optional, it happens either way)
```
#define SNES_PAD_ENABLED 1
#define SNES_PAD_DATA_PIN 22
#define SNES_PAD_LATCH_PIN 26
#define SNES_PAD_CLOCK_PIN 27
```

2. Play a while on e.g. the MiSTer SNES Core
3. preferably some mario romhack where a lot of button presses are required

NickGuyver:
fix for issue #484

clear all the dpad states after each poll, failed polls were preserving old dpad states

### DIFF
--- a/src/addons/snes_input.cpp
+++ b/src/addons/snes_input.cpp
@@ -84,6 +84,11 @@ void SNESpadInput::process() {
     if (nextTimer < getMillis()) {
         snes->poll();
 
+        dpadUp = false;
+        dpadDown = false;
+        dpadLeft = false;
+        dpadRight = false;
+
         uint16_t joystickMid = GAMEPAD_JOYSTICK_MID;
         if ( DriverManager::getInstance().getDriver() != nullptr ) {
             joystickMid = DriverManager::getInstance().getDriver()->GetJoystickMidValue();


arntsonl (approved):
Easy fix

PR #1709, opened 2026-09-10
fix issue 1535

### LINKED ISSUE #1535: Marco inputs occasionally doubling in Switch mode
### Expected Behavior

Macros should behave consistently between each loop of the macro

### Current Behavior

A programed macro will occasionally repeat an input, causing issues in situation that require a precise sequence of inputs. For example, a macro that is programmed to execute the inputs [plus, left, A, A] will occasionally go [plus, plus, left, A, A]

### Context

I was trying to make a macro to help me shiny hunt in the new Pokémon game on switch 2. In this game, it can be done simply by quick traveling to the same spot over and over, so all I need it to do is open the map, move the cursor slightly left, and then press the A button twice. 
The macro I made works about 70% of the time, but the rest of the time one or more inputs got doubled. For example, it would press plus to open the map, then immediately press plus again to close it, or move the cursor left to the quick travel location, then move left again, moving the cursor away from the desired location. 
I can't be certain, but it seems to happen after the "release and wait ___ frames". So a macro that would press A for 5 frames, wait for 30 frames, then press B for 5 frames, might instead press A for 5 frames, wait for 30 frames, press A for 5 frames, press B for 5 frames. Like I said though, I haven't done extensive testing of this. 

Version
Pico-W (GP2040-CE_0.7.11_PicoW.uf2)
Current: v0.7.11
Latest: v0.7.11
Raspberry pi w
https://www.amazon.com/dp/B0B72GV3K3

### Steps to Reproduce

1. create a new macro with activation mode "toggle"
2. enter the following into the "advanced" section
[{"buttonMask":512,"duration":83330,"waitDuration":333320},{"buttonMask":262144,"duration":83330,"waitDuration":333320},{"buttonMask":2,"duration":83330,"waitDuration":799968},{"buttonMask":2,"duration":83330,"waitDuration":3749850}]
3. run the macro in an environment where the inputs can be closely observed, preferably in Nintendo Switch mode, you should see inputs occasionally being repeated

### Screenshots & Files

> Please include any relevant screenshots, photos/recordings, or files here.

Here's a video of it working successfully once, then double-pressing plus

NickGuyver:
previous step was still selected after position incremented, moved to after the update
added return after reset()

### DIFF
--- a/src/addons/input_macro.cpp
+++ b/src/addons/input_macro.cpp
@@ -181,7 +181,6 @@ void InputMacro::runCurrentMacro() {
         return;
     }
 
-    MacroInput& macroInput = macro.macroInputs[macroInputPosition];
     Gamepad * gamepad = Storage::getInstance().GetGamepad();
     currentMicros = getMicro();
 
@@ -211,6 +210,7 @@ void InputMacro::runCurrentMacro() {
         if (macroInputPosition >= (macro.macroInputs_count)) {
             if ( macro.macroType == ON_PRESS ) {
                 reset(); // On press = no more macro
+                return;
             } else {
                 restart(macro); // On Hold-Repeat or On Toggle = start macro again
             }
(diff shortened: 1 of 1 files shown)


arntsonl (approved):
Nice easy macro fix

PR #1710, opened 2026-09-11
Formatting

Pelsin:
just ran: `npm run format`


arntsonl (approved):
Always down for some npm formatting

PR #1716, opened 2026-09-14
Remove unused translation strings

Pelsin:
Went through them manually, i think i got them all.
Might be a better way to verify this than navigating through the site


mizma:
I used to use ripgrep for each one last time I did a bit of this, maybe there is a better way to automate it but really couldn't come up with one...

arntsonl (approved):
Nice! saves some space for sure

PR #1717, opened 2026-09-15
gpio expansion fixes

NickGuyver:
fixes a few things related to the gpio expansion to 48 pins
swapped to 64-bit gpio masks and reads
animation storage was causing an oob write and failed saves 
webconfig available pins and pin previewer
swapped hardcoded adc pins for the new defines


arntsonl (approved):
Nice! hah I had a local build fixing this but this is cleaner.

